PR #52, opened 2022-12-14
Drop C++20 Requirement

### LINKED ISSUE #49: ✨ Only require C++17
### What's the problem this feature will solve?

At the moment C++20 is required to build QECC. The standard is still pretty new and it's adoption in compilers such as gcc and clang has been slow. As a result, the default version of gcc (9.4.) that comes with the last LTS version of Ubuntu (20.04) is not able to compile the project. While there are multiple solutions, such as upgrading the compiler or switching to the latest LTS version (22.04), this is not possible under all circumstances.

The library itself is not using any crazy new C++20 features. It merely uses the spaceship operator and things like `map.contains`. It should be relatively straight forward to lower the requirements of this library to C++17.

### Describe the solution you'd like

QECC should be compatible/compilable with C++17.

lucasberent:
## Description

Remove C++20 dependencies to lower requirements to C++17 standard.

Fixes #49 

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,8 @@ defaults:
 
 jobs:
   ubuntu-tests:
-    name: Tests ubuntu-22.04
-    runs-on: ubuntu-22.04
+    name: Tests ubuntu-20.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,8 @@ defaults:
 
 jobs:
   ubuntu-tests:
-    name: Tests ubuntu-20.04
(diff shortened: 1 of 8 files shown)


burgholzer (changes requested):
I think this is not the end of the line when it comes to C++20 features being used, e.g., 
https://en.cppreference.com/w/cpp/container/map/contains is a C++20 feature that is being used.

Maybe it helps to temporarily lower the GitHub Action runner image to `ubuntu-20.04` since it has an older compiler that might detect that.
Probably, enabling stricter warnings (https://stackoverflow.com/questions/31830848/how-to-make-g-force-me-to-adhere-to-c-standard) could certainly help.

Also, please make sure to fix the linter warnings.

burgholzer (approved):
LGTM 👍🏼 
Many thanks. We can always adapt later on, if we should find out that some C++20 features are still being used.

PR #101, opened 2023-03-30
✨ Added 4.8.8 color code

peter-janderks:
## Description
Adds a class ColorCode as generalization of the HexagonalColorCode to a color code with general functionality. 
Furthermore, a class SquareOctagonColorCode that can be used to generate the parity check matrix of the 4.8.8 triangular color code. 
Plotting is adapted for more general (color) codes with multiple logicals. 

## Checklist:


mqt/qecc/cc_decoder/test_square_octagon_color_code.py
burgholzer, line 1: This file should be part of the test directory. Otherwise the tests will never be executed.

mqt/qecc/cc_decoder/test_square_octagon_color_code.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+from square_octagon_color_code import SquareOctagonColorCode
+
+
+def test_d3():
+    d3_layout = SquareOctagonColorCode(3)
```
burgholzer, line 6: Similar to the other test file for the Hexagonal code, this should be a test fixture that is reused throughout the tests.

burgholzer, line 6: Similar to the other test file for the Hexagonal code, this should be a pytest fixture that is reused throughout the tests.

burgholzer, line 6: small nit: the `d3_` is somewhat redundant here. just call this parameter `layout` in all the tests. Probably solved by adopting the pytest fixture.

mqt/qecc/cc_decoder/test_square_octagon_color_code.py
```diff
@@ -0,0 +1,21 @@
...
+
+    assert d3_layout.data_qubits == {(0, 1), (2, 1), (6, 1), (3, 2), (5, 2), (3, 4), (5, 4)}
+    assert d3_layout.ancilla_qubits == {(4, 0), (1, 3), (4, 3)}
+
+
+def test_number_of_qubits():
+    for d in range(3, 16, 2):
```
burgholzer, line 13: I'd recommend making this a parametrised test instead of a loop over the distances.

mqt/qecc/cc_decoder/test_square_octagon_color_code.py
```diff
@@ -0,0 +1,21 @@
...
+        layout = SquareOctagonColorCode(d)
+        assert len(layout.data_qubits) == 1 / 2 * d**2 + d - 1 / 2
+        assert len(layout.ancilla_qubits) == (1 / 2 * d**2 + d - 1 / 2) // 2
+
+
+def test_H():
+    d3_layout = SquareOctagonColorCode(3)
+    assert (d3_layout.H == np.array([[0, 0, 1, 0, 1, 1, 1], [0, 1, 0, 1, 0, 1, 1], [1, 1, 1, 0, 0, 1, 0]])).all()
```
burgholzer, line 21: IIRC there is an `allClose` function particularly designed to compare numpy arrays. I would advise to use that here to make this a little more concise.

mqt/qecc/cc_decoder/test_square_octagon_color_code.py
burgholzer, line 1: There should also be some tests on decoding this new code (similar to the other test file). Depending on how the suggestions above are realized, we might be able to exploit some common synergies and actually combine both test files.

mqt/qecc/cc_decoder/square_octagon_color_code.py
burgholzer, line 1: This is a very general observation/comment/suggestion.

This file and the corresponding code construction actually shares a lot of the implementation with the existing 3.6.6 code. Would it be too hard to generalise the whole construction so that we don't need to repeatedly write (almost) the same code over and over?
There are multiple possible levels of generalisation:
 - it is definitely possible to combine the Triangular Hexagonal and the Square Ocatgonal Color Code generation. The code looks so similar, that this should be fairly straight-forward.
 - would it make sense to generalise the construction to x.2x.2x color codes? Is that a common construction?
 - would it make sense to generalise the construction to x.y.y color codes? Can the combination of shapes be arbitrary?

I'd say the higher we can set the abstraction level, the easier it is to use our code. Then, it should also be fairly simple to add generic routines for computing these L matrices for logical errors and such.

If it is not possible to generally describe the construction of the data and ancilla qubits, then we might still want to introduce a common base class that sort of unifies some of the construction, e.g., creating the PCM from the data and ancillary qubits or the L matrix, the faces_to_qubits dict, etc.
I believe a lot of the code can be shared.

It might even be, that a combination of both approaches above makes sense, i.e., a generic base class for color codes (or even more general), and then a derived class for like x.y.y codes.

burgholzer, line 1: This is probably the second major comment: the simulation and the plotting routines should be adapted in a fashion that they can handle the new kind of codes easily.
In the best case, if, tomorrow, someone implements a new color code version, that should be as straight-forward as possible and should not require many modifications. I think this PR is the perfect example for that.
We should be able to generalise in a way that makes additions and experimenting as easy as possible.

burgholzer (changes requested):
Many thanks for this PR.
The feedback, which you'll find below, is rather high level/conceptual. I am a bit hesitant to just merge this as it is, since my feeling is that lots of this PR is duplicate code or, at least, very similar to the Hexagonal ECC code. As you'll see from the comments, I really believe it is worth to invest some time here to generalise the concepts and unify the code.

Maybe @lucasberent can also help with that.

mqt/qecc/cc_decoder/color_code.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import TYPE_CHECKING
+
+import numpy as np
+from ldpc import mod2
+
+if TYPE_CHECKING:  # pragma: no cover
+    import numpy.typing as npt
+
+
+class LatticeType(Enum):
```
burgholzer, line 15: You should be able to make the string conversions a little easier with the following
```suggestion
class LatticeType(Enum, str):
```

mqt/qecc/cc_decoder/color_code.py
```diff
@@ -0,0 +1,61 @@
...
+        self.data_qubits: set[tuple[int, int]] = set()
+        self.qubits_to_faces: dict[int, list[int]] = {}
+        self.faces_to_qubits: dict[int, list[int]] = {}
+        self.lattice = lattice
+        self.add_qubits()
+        self.H = np.zeros((len(self.ancilla_qubits), len(self.data_qubits)), dtype=int)
+        self.construct_layout()
+        self.compute_logical()
+        self.n = len(self.qubits_to_faces)
+        self.k = self.L.shape[1]
```
burgholzer, line 38:
```suggestion
    def __init__(self, distance: int, lattice_type: LatticeType) -> None:
        """Initialize the color code."""
        self.distance = distance
        self.ancilla_qubits: set[tuple[int, int]] = set()
        self.data_qubits: set[tuple[int, int]] = set()
        self.qubits_to_faces: dict[int, list[int]] = {}
        self.faces_to_qubits: dict[int, list[int]] = {}
        self.lattice_type = lattice_type
        self.add_qubits()
        self.H = np.zeros((len(self.ancilla_qubits), len(self.data_qubits)), dtype=int)
        self.construct_layout()
        self.compute_logical()
        self.n = len(self.qubits_to_faces)
        self.k = self.L.shape[1]
```
Note that this will require further changes throughout.

mqt/qecc/cc_decoder/decoder.py
```diff
@@ -11,7 +11,9 @@
 import numpy as np
 from z3 import Bool, Not, Optimize, Xor, simplify
 
+from mqt.qecc.cc_decoder.color_code import ColorCode, LatticeType
 from mqt.qecc.cc_decoder.hexagonal_color_code import HexagonalColorCode
+from mqt.qecc.cc_decoder.square_octagon_color_code import SquareOctagonColorCode
```
burgholzer, line 16: These classes should all be importable from `cc_decoder`. You can accomplish that by adding the classes to the respective `__init__.py` file.

mqt/qecc/cc_decoder/decoder.py
```diff
@@ -155,40 +156,58 @@ def simulate_error_rate(distance: int, error_rate: float, nr_sims: int, solver_p
...
     """Run the decoding simulation for a given distance and error rate."""
-    data = simulate_error_rate(distance, error_rate, nr_sims, solver)
+    code: ColorCode
+    if lattice == str(LatticeType.HEXAGON.value):
+        code = HexagonalColorCode(distance)
+    elif lattice == str(LatticeType.SQUARE_OCTAGON.value):
+        code = SquareOctagonColorCode(distance)
+    else:
+        print("Unknown code lattice type: " + str(lattice))
+        return
```
burgholzer, line 207: You could extract that into a function that takes a string and returns a `ColorCode`.
That function could be placed in the `__init__.py` file for example.

With the changes to the enum, you should also not need these `str`, `value` constructs.

instead of just returning in case of an invalid code, this should throw an exception. probably `InvalidArgument`

mqt/qecc/cc_decoder/decoder.py
```diff
@@ -155,40 +156,58 @@ def simulate_error_rate(distance: int, error_rate: float, nr_sims: int, solver_p
...
+    data = simulate_error_rate(code, error_rate, nr_sims, solver)
     strg = solver.split("/")[-1]
-    filename = f"./distance={distance},p={round(error_rate, 4)},solver={strg}.json"
+    filename = f"./code={str(code.lattice.value)},distance={code.distance},p={round(error_rate, 4)},solver={strg}.json"
```
burgholzer, line 210:
```suggestion
    filename = f"./code={code.lattice_type},distance={code.distance},p={round(error_rate, 4)},solver={strg}.json"
```

mqt/qecc/cc_decoder/decoder.py
burgholzer, line 207: Untested suggestion:
```suggestion
    def get_code(lattice_type: LatticeType, distance: int) -> ColorCode:
        if lattice_type == LatticeType.HEXAGON:
            return HexagonalColorCode(distance)
        if lattice_type == LatticeType.SQUARE_OCTAGON:
            return SquareOctagonColorCode(distance)
       raise ValueError("Unknown lattice type: {lattice_type}")     
```

mqt/qecc/cc_decoder/plots.py
```diff
@@ -345,3 +358,6 @@ def main() -> None:
     else:
         msg = "Unknown mode: " + args.mode
         raise ValueError(msg)
+
+
+generate_plots(Path("/home/luca/Documents/codeRepos/mqt-qecc/qecc/results-4qb"), Path("./results.pdf"))
```
burgholzer, line 363: that statement shouldn't be here

test/python/test_cc_decoder.py
```diff
@@ -73,38 +81,22 @@ def test_maxsat_decoder(code: cc_decoder.HexagonalColorCode) -> None:
     assert est[1] == 1
 
 
-def test_simulate(distance: int, p: float, nr_sims: int) -> None:
+def test_simulate(d3_hexcode: HexagonalColorCode, distance: int, p: float, nr_sims: int) -> None:
```
burgholzer, line 84: the distance parameter and fixture isn't really needed anymore, right? it's all covered by the code parameter.

test/python/test_cc_decoder.py
```diff
@@ -114,23 +106,25 @@ def test_z3_solver(script_runner: ScriptRunner, distance: int, p: float, nr_sims
     assert ret.success
     assert not ret.stderr
 
-    result = check_and_load_json(f"distance={distance},p={round(p, 4)},solver=z3.json", results_dir)
+    result = check_and_load_json(
+        f"./code={d3_hexcode.lattice.value},distance={d3_hexcode.distance},p={round(p, 4)},solver=z3.json", results_dir
```
burgholzer, line 110:
```suggestion
        f"./code={d3_hexcode.lattice_type},distance={d3_hexcode.distance},p={round(p, 4)},solver=z3.json", results_dir
```

test/python/test_cc_decoder.py
```diff
@@ -157,18 +152,21 @@ def test_external_solver(
     assert ret.success
     assert not ret.stderr
 
-    result = check_and_load_json(f"distance={distance},p={round(p, 4)},solver={solver}.json", results_dir)
+    result = check_and_load_json(
+        f"./code={d3_hexcode.lattice.value},distance={d3_hexcode.distance},p={round(p, 4)},solver={solver}.json",
```
burgholzer, line 156:
```suggestion
        f"./code={d3_hexcode.lattice_type},distance={d3_hexcode.distance},p={round(p, 4)},solver={solver}.json",
```

test/python/test_cc_decoder.py
```diff
@@ -73,38 +81,22 @@ def test_maxsat_decoder(code: cc_decoder.HexagonalColorCode) -> None:
...
-
-    return result
-
-
-def test_z3_solver(script_runner: ScriptRunner, distance: int, p: float, nr_sims: int, results_dir: str) -> None:
+def test_z3_solver(
+    script_runner: ScriptRunner, d3_hexcode: HexagonalColorCode, p: float, nr_sims: int, results_dir: str
+) -> None:
     """Test the Z3 solver."""
+    d = d3_hexcode.distance
```
burgholzer, line 96: there is no real need for introducing that `d` here, right? might as well be removed. Holds for all the tests.

test/python/test_square_octagon_color_code.py
```diff
@@ -0,0 +1,100 @@
...
+def test_d3(code: SquareOctagonColorCode) -> None:
+    """Test coordinates of qubits for distance 3."""
+    assert code.data_qubits == {(0, 1), (2, 1), (6, 1), (3, 2), (5, 2), (3, 4), (5, 4)}
+    assert code.ancilla_qubits == {(4, 0), (1, 3), (4, 3)}
+
+
+@pytest.mark.parametrize("code", [3], indirect=True)
+def test_h(code: SquareOctagonColorCode) -> None:
+    """Test the parity check matrix for distance 3."""
+    assert np.array_equal(code.H, np.array([[0, 0, 1, 0, 1, 1, 1], [0, 1, 0, 1, 0, 1, 1], [1, 1, 1, 0, 0, 1, 0]]))
```
burgholzer, line 66: These could just use the `d3_so_code`

test/python/test_square_octagon_color_code.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@pytest.fixture()
+def results_dir() -> str:
+    """Return directory to store results."""
+    return "./results"
+
+
+@pytest.mark.parametrize("code", list(range(3, 23, 2)), indirect=True)
+def test_number_of_qubits(code: SquareOctagonColorCode) -> None:
```
burgholzer, line 50: I think you do not really need the fixture on top, but you can directly make a parametrized test out of the Code objects

burgholzer (changes requested):
Many thanks for the work on this. Looks much better now and much more general.
As most of the time, I have a couple of small comments on some things that I believe could be improved; see below. Most of them shouldn't take too long. Most of the suggestions should work out of the box, but some might require further adaptations. So please make sure to test locally.

One more thing: Please update the PR description accordingly.

burgholzer (approved):
LGTM

PR #236, opened 2024-07-01
Restrict numpy to version < 2.

pehamTom:
## Description

New numpy version break binary compatibility with Cython. For now we restrict numpy version until this issue is fixed.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
     "z3-solver>=4.11,<4.14",
     "qecsim",
     "ldpc>=0.1.50",
-    "numpy < 2",
+    "numpy < 2",  # some of our dependencies are not yet compatible with numpy 2.0
     "qiskit[qasm3-import]>=1.0.0",
     "qiskit-aer>=0.14.0",
     "bposd",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     "z3-solver>=4.11,<4.14",
     "qecsim",
     "ldpc>=0.1.50",
-    "numpy",
+    "numpy < 2",
```
burgholzer, line 43:
```suggestion
    "numpy < 2",  # some of our dependencies are not yet compatible with numpy 2.0
```

burgholzer (approved):
Just added a comment on the dependency upper bound so that we do not forget to remove it eventually.

This is also not per-se a problem of our package, but of one of our dependencies. I suspect the ldpc package, which has not yet released a version that is compatible with numpy 2.0.
We should bring this to their attention asap since establishing these kinds of upper caps in the community is a really bad thing.

pyproject.toml
pehamTom: Good point, thanks for adding the comment. I know it's a quick fix for now. But without this, the builds in my other PR fail.

PR #244, opened 2024-07-22
Automated Synthesis of Syndrome Extraction Circuits for CSS codes

pehamTom:
## Description

Finding optimal scheduling for the CNOTs in a syndrome extraction (SE) circuit is a computationally difficult scheduling task. This PR introduces methods for creating such schedules for any CSS code. 

Features:

- [x] Optimal synthesis of SE schedules for any CSS code instance.
- [x] Optimal synthesis of SE schedules for CSS code *families* based on regular plane tilings if the dependencies between qubits in the tiling are provided.
- [ ] Automated construction of dependencies given a tiling.

## Checklist:

- [ ] The pull request only contains commits that are related to it.
- [ ] I have added appropriate tests and documentation.
- [ ] I have made sure that all CI jobs on GitHub pass.
- [ ] The pull request introduces no new warnings and follows the project's style guidelines.

PR #246, opened 2024-07-23
Ftsp bench

pehamTom:
## Description

This PR adds circuit files and benchmark scripts for evaluating our fault-tolerant state preparation protocol.

Furthermore, this PR adds a module for working with codes in QECC.

## Checklist:


burgholzer:
I am fairly sure, that you do not want to distribute these files and benchmark scripts as part of the `mqt.qecc` package. They will bloat the binary distributions of the package without adding much value.
Hence, I would suggest to move these files out of the `src` folder and into a separate top-level `eval` or `scripts` or `<insert suitable name>`.
If you should move these files, please also adjust the linter config so that these files are also properly linted.

burgholzer:
Furthermore, this seems to contain some changes that are also in #244. Maybe updating the branch also here helps.

pehamTom:
> I am fairly sure, that you do not want to distribute these files and benchmark scripts as part of the `mqt.qecc` package. They will bloat the binary distributions of the package without adding much value. Hence, I would suggest to move these files out of the `src` folder and into a separate top-level `eval` or `scripts` or `<insert suitable name>`. If you should move these files, please also adjust the linter config so that these files are also properly linted.

Given that other subpackages of qecc also have evaluation scripts distributed with the sources I figured I would do the same. Also, given that synthesizing some of these circuits takes a long time I think it is best to provide some utility to load these circuits directly instead of recomputing them. If people want to use the circuits from our paper they can then directly use them from the python code.

The bloat is not very high since there are only a couple of small qasm files. Of course, it's a good idea to pull all evaluations into a top-level benchmark folder. But I would rather treat that as a separate issue.

pehamTom:
> Furthermore, this seems to contain some changes that are also in #244. Maybe updating the branch also here helps.

That is true. The other PR contains commits that are not related to it. I'll have to remove them from #244 once this is done.

burgholzer:
> > I am fairly sure, that you do not want to distribute these files and benchmark scripts as part of the `mqt.qecc` package. They will bloat the binary distributions of the package without adding much value. Hence, I would suggest to move these files out of the `src` folder and into a separate top-level `eval` or `scripts` or `<insert suitable name>`. If you should move these files, please also adjust the linter config so that these files are also properly linted.
> 
> Given that other subpackages of qecc also have evaluation scripts distributed with the sources I figured I would do the same. Also, given that synthesizing some of these circuits takes a long time I think it is best to provide some utility to load these circuits directly instead of recomputing them. If people want to use the circuits from our paper they can then directly use them from the python code.

Point taken. If this is part of the package functionality, it naturally makes sense to include the files. Excuse my ignorance.

> 
> The bloat is not very high since there are only a couple of small qasm files. Of course, it's a good idea to pull all evaluations into a top-level benchmark folder. But I would rather treat that as a separate issue.

👍🏼 Would you mind creating such an issue to keep track of that?
I guess it would be best to move all evaluation scripts to a top-level `eval` or (probably better) `scripts` directory (with potential subfolders if deemed necessary).

pehamTom:
Issue is created. Thanks for your input :+1:

lucasberent (approved):

PR #266, opened 2024-08-26
🏁 Windows Support

burgholzer:
## Description

This PR adds continuous testing and deployment for Windows.
To this end, it switches to the reusable MQT workflows for CI and CD.
Along the way, it fixes a couple of errors that only revealed themselves on Windows.

## Checklist:


burgholzer:
Alright. This is looking much better now already.

@pehamTom I just removed the timeouts from the state prep tests and that seems to have done it. Although some of these tests (sometimes) seem to take forever. Do you see any kind of way to reduce the tests or to speed them up?
Some of the (simulation) tests also sporadically fail because a certain threshold is not reached. Could you maybe fix the random seed for these kind of tests so that the results are reproducible and less flaky?

@lucasberent could you please look at the change here: https://github.com/cda-tum/mqt-qecc/pull/266/commits/b58b6045c8067d0850886a2549c0a178e12c59d7
This was necessary to fix one of the tests on Windows. However, the test now fails on Ubuntu and macOS. Do you see any kind of reason for that.
I was kind of suspicious that the following to lines are not the same despite the same comment:
- https://github.com/cda-tum/mqt-qecc/blob/63defae0fe7c2c4a8ea53cd932fd47c570214d7c/test/python/analog_info/test_memory_experiment.py#L111
- https://github.com/cda-tum/mqt-qecc/blob/63defae0fe7c2c4a8ea53cd932fd47c570214d7c/test/python/analog_info/test_memory_experiment.py#L137

pehamTom:
> @pehamTom I just removed the timeouts from the state prep tests and that seems to have done it. Although some of these tests (sometimes) seem to take forever. Do you see any kind of way to reduce the tests or to speed them up? Some of the (simulation) tests also sporadically fail because a certain threshold is not reached. Could you maybe fix the random seed for these kind of tests so that the results are reproducible and less flaky?

Well, the timeout is there, so the SAT solver doesn't spend a long time on unsatisfiable instances. I can try to trim down the tests further. But again, there are not many small code instances one can meaningfully test.

burgholzer:
> > @pehamTom I just removed the timeouts from the state prep tests and that seems to have done it. Although some of these tests (sometimes) seem to take forever. Do you see any kind of way to reduce the tests or to speed them up? Some of the (simulation) tests also sporadically fail because a certain threshold is not reached. Could you maybe fix the random seed for these kind of tests so that the results are reproducible and less flaky?
> 
> 
> 
> Well, the timeout is there, so the SAT solver doesn't spend a long time on unsatisfiable instances. I can try to trim down the tests further. But again, there are not many small code instances one can meaningfully test. 
> 
> 

At least on Windows that has led to basically none of the synthesis tasks returning a circuit (see the CI logs from a couple commits ago).
It's odd, but not completely unexpected that performance on windows is so different to the Unix systems.
Maybe relaxing the timeouts a little bit instead of removing them would be even better.
Just seeking a pragmatic solution here.
The current CI run shows that it works in principle given a long enough timeout.

lucasberent:
> Alright. This is looking much better now already.
> 
> @pehamTom I just removed the timeouts from the state prep tests and that seems to have done it. Although some of these tests (sometimes) seem to take forever. Do you see any kind of way to reduce the tests or to speed them up? Some of the (simulation) tests also sporadically fail because a certain threshold is not reached. Could you maybe fix the random seed for these kind of tests so that the results are reproducible and less flaky?
> 
> @lucasberent could you please look at the change here: [b58b604](https://github.com/cda-tum/mqt-qecc/commit/b58b6045c8067d0850886a2549c0a178e12c59d7) This was necessary to fix one of the tests on Windows. However, the test now fails on Ubuntu and macOS. Do you see any kind of reason for that. I was kind of suspicious that the following to lines are not the same despite the same comment:
> 
> * https://github.com/cda-tum/mqt-qecc/blob/63defae0fe7c2c4a8ea53cd932fd47c570214d7c/test/python/analog_info/test_memory_experiment.py#L111
> * https://github.com/cda-tum/mqt-qecc/blob/63defae0fe7c2c4a8ea53cd932fd47c570214d7c/test/python/analog_info/test_memory_experiment.py#L137

the test should be correct as is, the comment is wrong the estimate should be [1,0,0]..

burgholzer:
> > Alright. This is looking much better now already.
> > @pehamTom I just removed the timeouts from the state prep tests and that seems to have done it. Although some of these tests (sometimes) seem to take forever. Do you see any kind of way to reduce the tests or to speed them up? Some of the (simulation) tests also sporadically fail because a certain threshold is not reached. Could you maybe fix the random seed for these kind of tests so that the results are reproducible and less flaky?
> > @lucasberent could you please look at the change here: [b58b604](https://github.com/cda-tum/mqt-qecc/commit/b58b6045c8067d0850886a2549c0a178e12c59d7) This was necessary to fix one of the tests on Windows. However, the test now fails on Ubuntu and macOS. Do you see any kind of reason for that. I was kind of suspicious that the following to lines are not the same despite the same comment:
> > 
> > * https://github.com/cda-tum/mqt-qecc/blob/63defae0fe7c2c4a8ea53cd932fd47c570214d7c/test/python/analog_info/test_memory_experiment.py#L111
> > * https://github.com/cda-tum/mqt-qecc/blob/63defae0fe7c2c4a8ea53cd932fd47c570214d7c/test/python/analog_info/test_memory_experiment.py#L137
> 
> the test should be correct as is, the comment is wrong the estimate should be [1,0,0]..

In that case, something is going wrong in the decoder on Windows as it always yields `[0, 0, 0]`. Is there anything that you could think of that could cause this?
I'll revert the test change for now.

lucasberent:
> In that case, something is going wrong in the decoder on Windows as it always yields `[0, 0, 0]`. Is there anything that you could think of that could cause this? I'll revert the test change for now.

perhaps something with the np arrays? Maybe using 'astype' helps otherwise unclear to me.

lucasberent:
> > > In that case, something is going wrong in the decoder on Windows as it always yields `[0, 0, 0]`. Is there anything that you could think of that could cause this? I'll revert the test change for now.
> > 
> > 
> > perhaps something with the np arrays? Maybe using 'astype' helps otherwise unclear to me.
> 
> Nah, that can't help. The `res[0]` array is equal to `[0, 0, 0]`. So something in the decoder is producing a wrong result.
> 
> Btw: wtf went on with the last couple of commits? 😜

I have no idea i merged locally and suddenly had allthese changes in there, sorry if I broke anything :P

pehamTom:
> > > @pehamTom I just removed the timeouts from the state prep tests and that seems to have done it. Although some of these tests (sometimes) seem to take forever. Do you see any kind of way to reduce the tests or to speed them up? Some of the (simulation) tests also sporadically fail because a certain threshold is not reached. Could you maybe fix the random seed for these kind of tests so that the results are reproducible and less flaky?
> > 
> > 
> > Well, the timeout is there, so the SAT solver doesn't spend a long time on unsatisfiable instances. I can try to trim down the tests further. But again, there are not many small code instances one can meaningfully test.
> 
> At least on Windows that has led to basically none of the synthesis tasks returning a circuit (see the CI logs from a couple commits ago). It's odd, but not completely unexpected that performance on windows is so different to the Unix systems. Maybe relaxing the timeouts a little bit instead of removing them would be even better. Just seeking a pragmatic solution here. The current CI run shows that it works in principle given a long enough timeout.

Yeah this is pretty annoying. The tests that take the longest currently are the tests for the [[17,15]] code. But if I don't test that one, I have no test that checks whether our construction works for d=5 codes. The timeouts are really only there to make sure the tests don't take a really long time. By default the timeout is set to one hour. We can of course increase those, making the tests go even longer. 

I am not sure how to solve this except for writing tests that test the individual constraints in the construction. That would mean I'd have to write many more tests.

burgholzer:
> > > > @pehamTom I just removed the timeouts from the state prep tests and that seems to have done it. Although some of these tests (sometimes) seem to take forever. Do you see any kind of way to reduce the tests or to speed them up? Some of the (simulation) tests also sporadically fail because a certain threshold is not reached. Could you maybe fix the random seed for these kind of tests so that the results are reproducible and less flaky?
> > > 
> > > 
> > > Well, the timeout is there, so the SAT solver doesn't spend a long time on unsatisfiable instances. I can try to trim down the tests further. But again, there are not many small code instances one can meaningfully test.
> > 
> > 
> > At least on Windows that has led to basically none of the synthesis tasks returning a circuit (see the CI logs from a couple commits ago). It's odd, but not completely unexpected that performance on windows is so different to the Unix systems. Maybe relaxing the timeouts a little bit instead of removing them would be even better. Just seeking a pragmatic solution here. The current CI run shows that it works in principle given a long enough timeout.
> 
> Yeah this is pretty annoying. The tests that take the longest currently are the tests for the [[17,15]] code. But if I don't test that one, I have no test that checks whether our construction works for d=5 codes. The timeouts are really only there to make sure the tests don't take a really long time. By default the timeout is set to one hour. We can of course increase those, making the tests go even longer.
> 
> I am not sure how to solve this except for writing tests that test the individual constraints in the construction. That would mean I'd have to write many more tests.

Based on the offline discussions, I would say let's just increase the timeouts from the main branch just enough so that CI on Windows passes in most cases. It's ok if they turn out to be a little flaky as long as there is a high chance of the tests succeeding within 3 re-runs.

burgholzer:
> The Stateprep tests now go through with the new timeouts.

Many thanks for the adjustments! 🙏🏼

lucasberent:
fixed the test finally :+1:

burgholzer:
> fixed the test finally :+1: 

Are you sure? CI is still failing somehow 🥲

lucasberent:
numpy cap introduced due to dependency issues, see e.g., https://github.com/oscarhiggott/PyMatching/issues/102

burgholzer:
Ok. `pymatching 2.2.1` already puts a pin on `numpy < 2` so we don't have to.
Hopefully, all tests pass now and this can be merged. I'll then proceed to release a new version, which will be the first one to ship Windows wheels 🎉 

Also cleaned up the PR history a little bit so it becomes clearer what changed.

burgholzer:
@pehamTom sorry that I have to bother you again, but I can't seem to get the ft simulation tests to work under Windows 3.12; even after increasing the number of shots by a factor of 10.
The logical error rate seems to be quite off in the two ft simulation tests.
Windows 3.9 seems to be fine and is not complaining.
Any idea what could be causing this?

pehamTom:
> @pehamTom sorry that I have to bother you again, but I can't seem to get the ft simulation tests to work under Windows 3.12; even after increasing the number of shots by a factor of 10. The logical error rate seems to be quite off in the two ft simulation tests. Windows 3.9 seems to be fine and is not complaining. Any idea what could be causing this?

It was another timeout issue with the state preparation. Since some optimal circuits are synthesized for the simulation, they timed out and returned a non-verified circuit. 

But now the time to run the tests is 77 minutes on windows 3.12. I don't think that is an acceptable amount of time. I think the only way to circumvent this is to either not test the SAT solution for the state preparation and verification circuit synthesis or test the individual SAT formulas used in the encoding separately.

burgholzer:
> > @pehamTom sorry that I have to bother you again, but I can't seem to get the ft simulation tests to work under Windows 3.12; even after increasing the number of shots by a factor of 10. The logical error rate seems to be quite off in the two ft simulation tests. Windows 3.9 seems to be fine and is not complaining. Any idea what could be causing this?
> 
> 
> 
> It was another timeout issue with the state preparation. Since some optimal circuits are synthesized for the simulation, they timed out and returned a non-verified circuit. 
> 
> 
> 
> But now the time to run the tests is 77 minutes on windows 3.12. I don't think that is an acceptable amount of time. I think the only way to circumvent this is to either not test the SAT solution for the state preparation and verification circuit synthesis or test the individual SAT formulas used in the encoding separately.

Yeah. That's somewhat rough..
What about the most pragmatic solution of not running these tests under Windows but running them under all other operating systems?
It's not ideal, but at least we don't completely give up testing the functionality.

pehamTom:
> > > @pehamTom sorry that I have to bother you again, but I can't seem to get the ft simulation tests to work under Windows 3.12; even after increasing the number of shots by a factor of 10. The logical error rate seems to be quite off in the two ft simulation tests. Windows 3.9 seems to be fine and is not complaining. Any idea what could be causing this?
> > 
> > 
> > It was another timeout issue with the state preparation. Since some optimal circuits are synthesized for the simulation, they timed out and returned a non-verified circuit.
> > But now the time to run the tests is 77 minutes on windows 3.12. I don't think that is an acceptable amount of time. I think the only way to circumvent this is to either not test the SAT solution for the state preparation and verification circuit synthesis or test the individual SAT formulas used in the encoding separately.
> 
> Yeah. That's somewhat rough.. What about the most pragmatic solution of not running these tests under Windows but running them under all other operating systems? It's not ideal, but at least we don't completely give up testing the functionality.

That would be fine by me.

burgholzer:
Everything is passing and I am glad that this is finally in 🎉 
Nice step forward!

PR #272, opened 2024-08-28
Fix incorrect placement of z-errors

pehamTom:
## Description

Due to a missing check in the propagation of errors, some Z-errors were propagated incorrectly

## Checklist:


burgholzer (approved):
LGTM

Really just a comment because I stumbled upon it: we shouldn't be relying on private properties of Qiskit (`Bit._index`).
This is subject to change and access to the index (via the public `Bit.index`) is actually deprecated since way before Qiskit 1.0.

pehamTom:
I am currently refactoring code anyway since I am working on something new that overlaps a lot with the state prep functionality. Gonna get rid of those private accesses then.

PR #283, opened 2024-09-16
Fix error propagation.

pehamTom:
## Description

There is an error in the computation of fault sets in state preparation circuits. Sometimes, the gates are not traversed in the correct topological order, and thus, some faults are not correct. This didn't show up before because the faults were still caught by the verification circuits. 

Fixes # (issue)

## Checklist:


lucasberent (commented):
lgtm

PR #377, opened 2025-03-03
WIP: FT State Preparation Through Code Symmetries

inctechs:
## Description

We are trying to improve the fault-tolerant state preparation circuits by using code symmetries and/ or restructuring of circuits. 

## Checklist:

- [ ] The pull request only contains commits that are related to it.
- [ ] I have added appropriate tests and documentation.


docs/library/StatePrep.rst
pehamTom, line 1: This file has been removed from the documentation. Please remove it from this PR as well.

pyproject.toml
```diff
@@ -218,6 +218,7 @@ ignore = [
     "PLR2004", # Magic value used in comparison
     "PLC0415", # Import should be at top of file
     "S101",    # Use of assert detected
+    "T201",    # avoid removing print statements
```
pehamTom: I guess this is for development purposes. It shouldn't remain in the final version of the PR.

docs/StatePrep.ipynb
pehamTom, line 1: This file has been removed from the documentation. Please remove it from this PR as well.

src/mqt/qecc/analog_information_decoding/simulators/quasi_single_shot_v2.py
```diff
@@ -290,6 +290,6 @@ def run(self, samples: int = 1) -> dict[str, Any]:
             if run % self.save_interval == 1:
                 self._save_results(success_cnt, run)
                 if _check_convergence(success_cnt, run, self.code_params, self.eb_precision):
-                    print("Converged")  # noqa: T201
+                    print("Converged")
```
pehamTom: I guess pre-commit removed this because removing prints is disabled globally. The comment should remain in the final version of the PR so please double-check that this change is reverted before that.

src/mqt/qecc/analog_information_decoding/simulators/simulation.py
```diff
@@ -742,7 +742,7 @@ def run(self, samples: int) -> dict[str, Any]:
                     self.code_params,
                     self.eb_precision,
                 ):
-                    print("Result has converged.")  # noqa: T201
+                    print("Result has converged.")
```
pehamTom: I guess pre-commit removed this because removing prints is disabled globally. The comment should remain in the final version of the PR so please double-check that this change is reverted before that.

src/mqt/qecc/circuit_synthesis/__init__.py
```diff
@@ -29,8 +30,11 @@
     "DeterministicVerification",
     "DeterministicVerificationHelper",
     "LutDecoder",
+    "NoisyDFTStatePrepSimulator",
```
pehamTom: The `DFTStatePrepSimulator` has been removed recently due to issues with qsample. Make sure to remove the export here.

src/mqt/qecc/circuit_synthesis/encoding.py
```diff
@@ -51,10 +55,13 @@ def heuristic_encoding_circuit(
...
-        np.vstack((checks, logicals)),
-        parallel_elimination=optimize_depth,
-    )
+    ge = GaussianElimination(matrix=np.vstack((checks, logicals)), code=code, parallel_elimination=optimize_depth)
+    ge.basic_elimination()
+    checks, cnots = ge.matrix, ge.eliminations
+    # checks, cnots = heuristic_gaussian_elimination(
+    #     np.vstack((checks, logicals)),
+    #     parallel_elimination=optimize_depth,
+    # )
```
pehamTom: Remove comment.

pyproject.toml
```diff
@@ -156,7 +156,7 @@ exclude = [
 
 [[tool.mypy.overrides]]
 module = ["qiskit.*", "qecsim.*", "qiskit_aer.*", "matplotlib.*", "scipy.*", "ldpc.*", "pytest_console_scripts.*",
-    "z3.*", "bposd.*", "numba.*", "pymatching.*", "stim.*", "multiprocess.*", "sinter.*", "qsample.*", "pandas.*"]
+    "z3.*", "bposd.*", "numba.*", "pymatching.*", "stim.*", "multiprocess.*", "sinter.*", "qsample.*", "pandas.*", "sympy.*"]
```
pehamTom: After internal discussions, please remove.

src/mqt/qecc/codes/css_code.py
```diff
@@ -80,6 +81,25 @@ def __init__(
...
+    def automorphism_group(self, generators: list[list[int]]) -> None:
+        self._automorphism_generators = [Permutation(generator) for generator in generators]
+        g = PermutationGroup(self._automorphism_generators)
+        self._automorphism_group = list(g.generate())
+
+    @property
+    def automorphism_generators(self) -> list[Permutation]:
+        """Property for the automorphism generators."""
+        return self._automorphism_generators
+
```
pehamTom: Given that this is not used anywhere, I would remove it. If the functionality is needed, we can add it then.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+class CandidateAction(Enum):
+    """Class to help distinguish the control flow of the reference based heuristic gaussian elimination."""
+
+    SKIP = auto()  # Ignore this candidate and continue the loop
+    RESTART_SEARCH = auto()  # Reset used_columns and break to restart the search
+    TRIGGER_BACKTRACK = auto()  # Call backtrack() and break
+    EVALUATE = auto()  # Proceed with full validation (overlap checks)
+
+
+class GaussianElimination:
```
pehamTom: Thinking about this, maybe generically calling this class "GaussianElimination" is a bit misleading. The functionality here is very specialized for circuit synthesis. Somebody else looking at the code might be very surprised to find fault sets and codes as parameters for a class that seems to only do Gaussian elimination. 

Open to suggestions, but what about "EliminationCNOTSynthesis" (or EliminationCNOTSynthesizer)?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+    """Class to apply Gaussian Elimination on a given Matrix."""
```
pehamTom: Elaborate more about the functionality of the class in the Docstring.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        self.used_cnots: list[tuple[int, int]] = []
+
+    def basic_elimination(self) -> None:
+        """Basic heuristic Gaussian elimination.
+
+        Calculates CNOTS and Check matrix.
+        """
+        while not self.is_reduced():
+            costs_unused = self._mask_out_used_qubits()
+            if self._handle_stagnation(costs_unused):
```
pehamTom: Writing this here, since it is the first context where this comes up. The name `_handle_stagnation` is not clear to me. What about "_reset_if_stuck"?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+                self.used_columns = []
+            return True
+        return False
+
+    def _triangular_reset(self) -> None:
+        logger.warning("Local minimum reached. Making matrix triangular.")
+        self.matrix = mod2.row_echelon(self.matrix, full=True)[0]
+        self.costs = self._compute_cost_matrix()
+
+    def _apply_cnot_to_matrix(self, i: int, j: int) -> None:
```
pehamTom, line 376: Consider changing the types of i and j since you always perform a type cast when using this function.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        Calculates CNOTS and Check matrix.
+        """
+        while not self.is_reduced():
+            costs_unused = self._mask_out_used_qubits()
+            if self._handle_stagnation(costs_unused):
+                continue
+            i, j = np.unravel_index(np.argmin(costs_unused), self.costs.shape)
+            self._apply_cnot_to_matrix(int(i), int(j))
+
+    def reference_based_construction(self) -> None:
```
pehamTom: We called this fault-set-guided synthesis in the paper. Should we use the same here?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        if self.guide_by_x:
+            self.failed_cnots = [fcnot for fcnot in self.failed_cnots if removed_cnot[0] != fcnot[0]]
+        else:
+            self.failed_cnots = [fcnot for fcnot in self.failed_cnots if removed_cnot[1] != fcnot[1]]
+        self.failed_cnots.append(removed_cnot)
+        # print_dynamic_eliminations(eliminations, failed_cnots)
+        self.backtrack_required = False
+
+    def _validate_inputs(self) -> None:
+        """Here will be some checks that ensure that the reference based construction can actually be executed."""
```
pehamTom: Why "will be"? Aren't they here?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        has_refs = bool(self.ref_x_fs.size or self.ref_z_fs.size)
+        has_code = self.code is not None
+        if has_refs ^ has_code:  # xor
```
pehamTom: Why bit-wise XOR instead of `!=`?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+                npt.NDArray[np.int8],
+                list[tuple[int, int]],
+                list[tuple[int, int]],
+            ]
+        ] = []
+        self.current_x_fs: npt.NDArray[np.int8] = np.eye(self.matrix.shape[1], dtype=np.int8)
+        self.current_z_fs: npt.NDArray[np.int8] = np.eye(self.matrix.shape[1], dtype=np.int8)
+        self.used_cnots: list[tuple[int, int]] = []
+
+    def basic_elimination(self) -> None:
```
pehamTom: Given that the name of the class is changed to something that indicates that it is responsible for synthesizing CNOT circuits, maybe we should call this "greedy_synthesis" or something along those lines. See also my comment below about the reference-based elimination.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        np.fill_diagonal(costs, 1)
+        return costs
+
+    def _mask_out_used_qubits(self) -> npt.NDArray[np.int8]:
+        m = np.zeros((self.matrix.shape[1], self.matrix.shape[1]), dtype=bool)  # type: npt.NDArray[np.bool_]
+        m[self.used_columns, :] = True
+        m[:, self.used_columns] = True
+        return np.ma.array(self.costs, mask=m)  # type: ignore[no-untyped-call]
+
+    def _modify_matrix_structure(self) -> None:
```
pehamTom: The name is too general. This method is only used in one place anyway. I would suggest doing the distance check above and then just calling `row_echelon`

General point of discussion about this method. As far as I understand, you do this normalization to make finding a solution more reliable. But finding a solution based on the input matrix might still be possible, right? I think we can make the fact that the synthesis can depend on the matrix transparent. If the method fails to give a solution the caller can then manually call the normalization function before retrying the synthesis.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+    def reference_based_construction(self) -> None:
+        """Reference based heuristic Gaussian elimination.
+
+        This version takes reference fault sets into consideration and only applies cnots that
+        do not cause an overlap of the reference fault set and the fault set of the newly
+        created circuit.
+        """
+        self._validate_inputs()
+        self._modify_matrix_structure()
+        self._ref_based_init()
```
pehamTom: All data related to the reference-based synthesis is initialized here. I think we could then also move the arguments needed for this synthesis into this method, instead of having all the reference fault-sets passed via the `__init__` method. Do you see a downside to that?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+                    if is_valid:
+                        self.used_cnots.append((int(i), int(j)))
+                        self._commit_to_cnot(i, j, new_errors)
+                        self._apply_cnot_to_matrix(int(i), int(j))
+                        break
+                    self.failed_cnots.append((int(i), int(j)))
+                    continue
+
+    def is_reduced(self) -> bool:
+        """Method decides if the Gaussian elimination has successfully ended."""
```
pehamTom: I know that I also didn't do that, but maybe it would be good to explain what the condition is since it is hard to see from the code.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        """
+        self._validate_inputs()
+        self._modify_matrix_structure()
+        self._ref_based_init()
+        while not self.is_reduced():
+            costs_unused = self._mask_out_used_qubits()
+            if self._handle_stagnation(costs_unused):
+                continue
+            candidate_pairs = self._get_candidate_pairs(costs_unused)
+            for i, j in candidate_pairs:
```
pehamTom, line 220: Since there are a lot of nested conditions, it might be an idea to separate this into another function just to make the flow very clear. It is not necessary, however, I think. These kind of backtracking approaches are inadvertently convoluted.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+                    self.used_columns = []
+                    self.backtrack_required = True
+                    break
+
+                if action == CandidateAction.TRIGGER_BACKTRACK:
+                    if self.stack:
+                        self._backtrack()
+                    else:
+                        # This case means we backtracked to the beginning and still failed.
+                        logger.error("Backtracking stack is empty. No solution found.")
```
pehamTom, line 234: Is this really an error? Finding no solution is a possible outcome of the synthesis. I would just log it as "INFO". You could also have the method return a Boolean indicating whether the synthesis succeeded or not.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+    """
+    # NOTE: This implementation assumes single error events only. For higher-order errors, adjustments are required.
+    control, target = cnot_gate
+
+    control_row = propagation_matrix[control]
+    target_row = propagation_matrix[target]
+
+    if x_error:
+        propagation_matrix[control] = np.bitwise_xor(control_row, target_row)
+        return propagation_matrix[control], propagation_matrix
```
pehamTom, line 470:
```suggestion
    update_row = control if x_error else target
    propagation_matrix[update_row] = np.bitwise_xor(control_row, target_row)
    return propagation_matrix[update_row], propagation_matrix
```

Simplifies the logic a little bit.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
 
-    The algorithm utilizes a greedy heuristic to select the columns to eliminate in order to minimize the number of eliminations required.
+    propagation_matrix[target] = np.bitwise_xor(control_row, target_row)
+    return propagation_matrix[target], propagation_matrix
```
pehamTom, line 473:
```suggestion
```

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        m[:, self.used_columns] = True
+        return np.ma.array(self.costs, mask=m)  # type: ignore[no-untyped-call]
+
+    def _modify_matrix_structure(self) -> None:
+        """This should not necessary but for distance seven codes this was the only way to reliably produce solutions."""
+        if self.code.distance > 5:
+            if self.ref_z_fs.size and not self.ref_x_fs.size:
+                self.matrix = mod2.row_echelon(self.matrix, full=True)[0]
+            if self.ref_z_fs.size and self.ref_x_fs.size:
+                self.matrix = mod2.row_echelon(self.matrix, full=True)[0]
```
pehamTom:
```suggestion
        if self.code.distance > 5 and self.ref_z_fs.size:
                self.matrix = mod2.row_echelon(self.matrix, full=True)[0]
```

Am I overlooking something or can this be simplified? The same code is executed no matter what `self.ref_x_fs.size` is.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+
+    def _get_candidate_pairs(self, costs_unused: npt.NDArray[np.int8]) -> list[tuple[int, int]]:
+        # Get all valid (i, j) pairs sorted by cost
+        candidate_indices = np.argsort(costs_unused.flatten())  # Flatten and sort by value
+        return [(int(i), int(j)) for idx in candidate_indices for i, j in [np.unravel_index(idx, self.costs.shape)]]
+        # return [np.unravel_index(idx, self.costs.shape) for idx in candidate_indices]
```
pehamTom, line 408:
```suggestion
```

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+    def _check_overlap(
+        self,
+        ref_fs: npt.NDArray[np.int8],
+        ref_1_fs: npt.NDArray[np.int8],
+        current_fs: npt.NDArray[np.int8],
+        new_error: npt.NDArray[np.int8],
+        x_error: bool = True,
+    ) -> bool:
+        overlap: bool = True
+        if ref_fs.size:
```
pehamTom, line 419: Can this ever not be the case?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+            overlap = self.code.check_fs_overlap(ref_fs, new_error, x_error=x_error)
+            if ref_1_fs.size and not overlap:
+                trial_2_fs = (current_fs + new_error) % 2
```
pehamTom, line 422: You used `^` everywhere else throughout (thanks for correcting my usage of `%2` by the way). Is there a reason it is different here?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        # return [np.unravel_index(idx, self.costs.shape) for idx in candidate_indices]
+
+    def _check_overlap(
+        self,
+        ref_fs: npt.NDArray[np.int8],
+        ref_1_fs: npt.NDArray[np.int8],
+        current_fs: npt.NDArray[np.int8],
+        new_error: npt.NDArray[np.int8],
+        x_error: bool = True,
+    ) -> bool:
```
pehamTom, line 417: Make sure to add a docstring explaining what happens here. Just looking at the code it is hard to say what happens here. Not a criticism of the code, but this is a crucial step of the algorithm and we want to be clear about what happens here.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+            overlap = self.code.check_fs_overlap(ref_fs, new_error, x_error=x_error)
+            if ref_1_fs.size and not overlap:
+                trial_2_fs = (current_fs + new_error) % 2
+                trial_2_fs = self._filter_fault_set(trial_2_fs, int((self.code.distance - 1) / 2))
+                overlap = self.code.check_fs_overlap(ref_1_fs, trial_2_fs, x_error=x_error)
+        return overlap
+
+    @staticmethod
+    def _check_error_existence(error: npt.NDArray[np.int8], error_memory: set[npt.NDArray[np.int8]]) -> bool:
+        # INFO: Quickly check if new error is known to cause overlap. This saves another long overlap check.
```
pehamTom, line 429:
```suggestion
        """Quickly check if the error is known to cause overlap. This saves another long overlap check."""
```

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        if ref_fs.size:
+            overlap = self.code.check_fs_overlap(ref_fs, new_error, x_error=x_error)
+            if ref_1_fs.size and not overlap:
+                trial_2_fs = (current_fs + new_error) % 2
+                trial_2_fs = self._filter_fault_set(trial_2_fs, int((self.code.distance - 1) / 2))
+                overlap = self.code.check_fs_overlap(ref_1_fs, trial_2_fs, x_error=x_error)
+        return overlap
+
+    @staticmethod
+    def _check_error_existence(error: npt.NDArray[np.int8], error_memory: set[npt.NDArray[np.int8]]) -> bool:
```
pehamTom: To use the commonly used terminology, I would call the `error_memory` an `error_cache`.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        # INFO: Quickly check if new error is known to cause overlap. This saves another long overlap check.
+        error_tuple = tuple(error.flatten())
+        return error_tuple in error_memory
+
+    @staticmethod
+    def _filter_fault_set(fault_set: npt.NDArray[np.int8], threshold: int) -> npt.NDArray[np.int8]:
+        """Filter fault sets based on a sum threshold."""
```
pehamTom, line 435:
```suggestion
        """Filter faults based on weight."""
```

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+#     sys.stdout.flush()
+
+
+class CandidateAction(Enum):
+    """Class to help distinguish the control flow of the reference based heuristic gaussian elimination."""
+
+    SKIP = auto()  # Ignore this candidate and continue the loop
+    RESTART_SEARCH = auto()  # Reset used_columns and break to restart the search
+    TRIGGER_BACKTRACK = auto()  # Call backtrack() and break
+    EVALUATE = auto()  # Proceed with full validation (overlap checks)
```
pehamTom, line 102: Since this class is only used internally by the `GaussianElimination` class, we can make it private to that class.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+        code: CSSCode,
+        parallel_elimination: bool = True,
+        ref_x_fs: npt.NDArray[np.int8] | None = None,
+        ref_z_fs: npt.NDArray[np.int8] | None = None,
+        ref_x_1fs: npt.NDArray[np.int8] | None = None,
+        ref_z_1fs: npt.NDArray[np.int8] | None = None,
+        penalty_cnots: list[tuple[int, int]] | None = None,
+        guide_by_x: bool = True,
+    ) -> None:
+        """Initialiser for the basic functionality."""
```
pehamTom: It is important here to have a clear description of the arguments.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+            self.x_propagation_matrix.copy(),
+            self.z_propagation_matrix.copy(),
+            self.used_columns.copy(),
+            self.matrix.copy(),
+            self.costs.copy(),
+            self.used_cnots.copy(),
+            self.failed_cnots.copy(),
+        ))
+        if self.guide_by_x:
+            self.failed_cnots = [fcnot for fcnot in self.failed_cnots if int(i) != fcnot[0]]
```
pehamTom, line 325: Why the cast to int here?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+    Take list of SPCs with already mutually disjoint fault sets and another list of potential candidates for new SPCs that
+    have a mutually disjoint fault set with all current SPCs.
 
     Args:
-        matrix: The matrix to perform Gaussian elimination on.
-        parallel_elimination: Whether to prioritize elimination steps that act on disjoint columns.
+        c_spcs: current SPCs that are already mutually disjoint in terms of fault sets
+        p_spcs: potential SPCs that shall be tested for mutually disjointness against the current set
+        x_error: flag that decides if check happens for x errors or z errors
+    """
```
pehamTom, line 488: Missing return value description.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+    i = 0
+    while i < len(p_spcs):
+        pspc = p_spcs[i]
```
pehamTom, line 491:
```suggestion
    for i, pspc in enumerate(p_spcs):
```

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -90,73 +94,452 @@ def iterative_search_with_timeout(
...
+            elif _spc.code.check_fs_overlap(z_fs, pz_fs, x_error=False):
+                break  # Stop checking if there's an overlap
+        else:  # No overlap found, so add pspc
+            c_spcs.append(pspc)
+
+        i += 1  # Move to next element in p_spcs
+    return c_spcs
+
+
+def get_fs_based_on_d(spc: StatePrepCircuit) -> tuple[npt.NDArray[np.int8], npt.NDArray[np.int8]]:
```
pehamTom: I am realizing more and more that dealing with fault sets should be better supported since it is an essential primitive in a lot of our synthesis tasks. Do you think it makes sense to add more support for fault sets? Nothing for this PR but maybe we can open an issue and deal with that somewhat soon.

Regarding this method specifically, I would argue that the calling context knows the distance. The logic between the d=3,5,7 cases are different enough that you could have a dedicated method for each and avoid the branching here.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -322,7 +705,8 @@ def build_css_circuit_from_cnot_list(n: int, cnots: list[tuple[int, int]], hadam
         The quantum circuit.
     """
     circ = QuantumCircuit(n)
-    circ.h(hadamards)
+    if hadamards:
```
pehamTom, line 726: Does qiskit not handle the case of empty arguments gracefully?

src/mqt/qecc/ecc_qiskit_wrapper.py
pehamTom, line 1: This file has been removed from QECC. Please remove it.

test/python/circuit_synthesis/test_gaussian_elimination.py
```diff
@@ -0,0 +1,274 @@
...
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from mqt.qecc.circuit_synthesis.synthesis_utils import CandidateAction, GaussianElimination
+from mqt.qecc.codes.css_code import CSSCode
+
+STEANE_CHECK_MATRIX = np.array([[1, 1, 0, 0, 1, 1, 0], [1, 0, 1, 0, 1, 0, 1], [0, 0, 0, 1, 1, 1, 1]], dtype=np.int8)
+STEANE_CODE = CSSCode(Hx=STEANE_CHECK_MATRIX)
```
pehamTom, line 25: Since these are only used in one fixture, I would define them in the fixture. 

Also you can get the code with `CSSCode.from_code_name("steane")`

test/python/circuit_synthesis/test_gaussian_elimination.py
```diff
@@ -0,0 +1,274 @@
...
+)
+def test_get_candidate_action(get_instance, cost, backtrack_required, used_cnots, failed_cnots, expected_action):
+    """Verify that _get_candidate_action returns the correct action based on state."""
+    # Arrange: Set up the instance state for the test case
+    get_instance.backtrack_required = backtrack_required
+    get_instance.used_cnots = used_cnots
+    get_instance.failed_cnots = failed_cnots
+
+    # Act: Call the method
+    action = get_instance._get_candidate_action(i=0, j=1, costs=cost)
```
pehamTom, line 60: This applies to other parts of this file as well. Testing private methods is generally considered bad practice for a variety of reasons. I also think we don't need to test on this level of granularity, since we do coverage tests anyway that show whether the different scenarios or candidate actions are executed. 

You obviously have more experience with your implementation. Do you think it is possible to write test cases based on the "matrix" level that test the different scenarios?

test/python/circuit_synthesis/test_gaussian_elimination.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Assert: Check the result
+    assert action == expected_action
+
+
+def test_apply_cnot_to_matrix_updates_matrix_and_costs_correctly(get_instance):
+    """Verify the target column of the matrix is correctly updated after a CNOT."""
+    expected_matrix = np.array([[1, 1, 0, 0, 0, 1, 0], [1, 0, 1, 0, 0, 0, 1], [0, 0, 0, 1, 1, 1, 1]], dtype=np.int8)
+    get_instance._apply_cnot_to_matrix(i=0, j=4)
```
pehamTom, line 69: As above.

test/python/circuit_synthesis/test_gaussian_elimination.py
```diff
@@ -0,0 +1,274 @@
...
+    expected_costs = expected_instance.costs
+    np.testing.assert_array_equal(get_instance.costs, expected_costs)
+
+
+def test_mask_out_used_qubits(get_instance):
+    """Check that the matrix is masked out cleanly based on the used columns."""
+    used_qubit_index = 1
+    unused_qubit_index = 0
+    get_instance.used_columns.append(used_qubit_index)
+    masked_costs = get_instance._mask_out_used_qubits()
```
pehamTom, line 81: Same as above.

pehamTom (changes requested):
Thanks for reworking the logic of the CNOT synthesis part. I have left some comments throughout that hopefully help to improve the code. 

I think the part that requires the most refactoring is the tests. I have written a comment inline but to give more justification why I think it is not a good idea here to test private methods. Testing that the synthesis works correctly, should be independent of the class internals. We want to test whether the public interface of the class works correctly. I agree that this case is a bit difficult because there is a lot of logic hidden in the `GaussianElimination` class. I hope that we can find a way to formulate good synthesis cases without accessing the class internals. The matrices do not have to make sense from an error correction point of view anyway. Let me know if I can help with that.

Before I forget, the documentation also has to be done.

src/mqt/qecc/analog_information_decoding/simulators/analog_tannergraph_decoding.py
```diff
@@ -254,7 +254,7 @@ def run(self, samples: int) -> dict[str, Any]:
                     self.code_params,
                     self.eb_precision,
                 ):
-                    print("Result has converged.")  # noqa: T201
+                    print("Result has converged.")
```
inctechs: Also fixed this as the removed comments below

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
inctechs: "Will be", as I was not sure if those were all the checks necessary. In the original version, there were a lot of checks to ensure that the calculations could actually be done and I did not know which I will need to transfer.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
inctechs: Yes, finding a solution should be possible without, but the algorithm starts too close to the original matrix and hence takes too long to find a solution. Therefore, this modification was the only way to 'quickly' find solutions for distance 7 codes then.

But yes I generally like the idea of leaving it out and up to the user to plug in a row echelon reduced matrix if necessary.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
inctechs: Yeah, so the idea of the candidate pair action enums was to make the control flow as clear as possible. Having the while-loop as the driving force of the algorithm and then the enums as different control flows for each iteration of the while loop, basically. So, I am open to making it clearer as this was also my objective, but I do not yet see how a different function would change that more?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
inctechs: The Elimination part would be crucial as there could be other 'CNOTSynthesizer'?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
inctechs: and then initialize them here as class variables?, so inside _ref_based_init() set self.ref_x_fs, etc.?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
inctechs: learned something new and wanted to use it.
It is corrected now together with a logical error in this function

src/mqt/qecc/ecc_qiskit_wrapper.py
inctechs, line 1: should this file be deleted? I read it somewhere else in a comment from you @pehamTom

src/mqt/qecc/circuit_synthesis/state_prep.py
```diff
@@ -293,6 +345,346 @@ def gate_optimal_prep_circuit(
     return FaultyStatePrepCircuit(circ, code.x_distance // 2, code.z_distance // 2)
 
 
+def iter_cyclic_combinations(lists):
```
inctechs: In this section, there are still some mypy issues. Should I go over this, or do you want to do so @pehamTom, as you can better judge which of the following is still needed and which is not in use anymore?

PR #388, opened 2025-03-31
Lattice Surgery Compilation

LSHerzog:
## Description

Adds features to perform macroscopic lattice surgery compilation for the color code. Both a routing for CNOT+T gates as well as a hill climbing based qubit label optimization is included. Furthermore, the microscopic construction of the snakes' stabilizers (for long distance lattice surgery) is provided both for the color code and surface code substrate.

## Checklist:

- [x] I have added appropriate tests and documentation. (At least a few tests and a rst file with the most important information)


docs/Co3.rst
pehamTom, line 1: There are both the `rst` and the `md` file in the docs now. I guess the `rst` file can be removed?

docs/co3.md
```diff
@@ -0,0 +1,37 @@
+---
+file_format: myst
+---
+
+# Compilation beyond the Surface Code `co3`
+
+This submodule contains an elementary routing routine for CNOT + T compilation on a hexagonal routing graph.
+Moreover the routing assumes that each accessible boundary can host both Z and X operators for lattice surgery.
+Hence, this code is valid for color codes only. Adaptions needed to incorporate valid paths for e.g. the folded surface code substrate.
```
pehamTom: How hard is it to adapt this? Does it make sense to have a section about how to potentially extend this?

docs/co3.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Layouts
+
+Basic layouts (sparse, pair, row, hex) can be generated automatically in the `HexagonalLattice` class.
```
pehamTom: Now that the autoapi correctly generates the source documentation, it would be cool if this links to the respective classes. It is not a must, though. If you want to change it, then please also change the rest of the method and function references accordingly.
```suggestion
Basic layouts (sparse, pair, row, hex) can be generated automatically in the [HexagonalLattice](co3.utils.lattice_router.HexagonalLattice) class.
```

docs/co3.md
pehamTom, line 1: I think that a guided showcase of how to use the code might be beneficial here. Is there a way to compile a minimal example step by step? Or are the instances too large to have a nice visual?

scripts/co3/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    "collect_data_space_time",
+    "gen_layout",
+    "plot_f_vs_t",
+    "plot_f_vs_t_subfigs",
+    "plot_improvement_circuit_types",
+    "plot_improvement_f_variation",
+    "plot_ratio_vs_t",
+    "plot_space_time",
+    "remove_edge_per_factory",
+]  # Controls what gets imported when using 'from co3.plots import *'
```
pehamTom:
```suggestion
]  
```

src/mqt/qecc/co3/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+    "SnakeBuilderSC",
+    "SnakeBuilderSTDW",
+    "compare_original_dynamic_gate_order",
+    "generate_max_parallel_circuit",
+    "generate_min_parallel_circuit",
+    "generate_random_circuit",
+    "translate_layout_circuit",
+]
+
+# __all__ += ["plots"]
```
pehamTom: Seems like a leftover line from when the plotting was part of src.
```suggestion

```

src/mqt/qecc/co3/microscopic/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2023 - 2025 Chair for Design Automation, TUM
+# All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+"""TODO."""
```
pehamTom: I added this `__init__` file because otherwise, autodoc does not document the code in this folder. For some reason, it needs a module description to work properly. I didn't know what a good description would be here. Could you give a brief submodule description here, please?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+from mqt.qecc import CSSCode
+
+
+class SnakeBuilderSC:
+    """Constructs a n-snake of distance d with a surface code in snake shape on the square lattice substate (leading to brickwall routing graph aka hex graph)."""
+
+    def __init__(
+        self,
+        g: nx.Graph,
+        positions_rough: list[list[tuple[int, int]]],
```
pehamTom: Instead of `list[list[tuple[int, int]]]`, is there a more descriptive type definition?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        positions_smooth: list[list[tuple[int, int]]],
+        d: int,
+    ) -> None:
+        """Initializes a SC n-snake.
+
+        Args:
+            g (nx.Graph): quadratic graph
```
pehamTom, line 40: Would it make sense to take the dimension of the grid as a parameter and construct the graph in the `__init__` method? It would make this class safer to use. But I also think, writing the requirement to the graph into the argument description is sufficient as well.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+            positions_rough (list[list[list[tuple[int,int]]]]): Node positions constituting the rough boundaries
+            positions_smooth (list[list[list[tuple[int,int]]]]): Node positions constituting the smooth boundaries
+            d (int): distance
+        """
+        self.g = g
+        self.positions_rough = positions_rough
+        self.positions_smooth = positions_smooth
+        self.d = d
+        assert len(positions_smooth) == 2, "To encode 1 logical qubit there must be 2 smooth boundaries."
+        assert len(positions_rough) == 2, "To encode 1 logical qubit there must be 2 rough boundaries."
```
pehamTom: What is the rationale for using assertions here? Since the positions are a parameter, it is better to handle this with exceptions.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        dist = (
+            min([len(el) for el in positions_rough] + [len(el) for el in positions_smooth]) - 1
+        )  # -1 because we count edges not nodes
+        assert dist == d, f"Distance d={dist} does not coincide with the geometry of the rough and smooth positions."
+
+        # must ensure that given G is big enough to place the stabilizers on it (maybe remove G from input and generate it within the class)
```
pehamTom, line 62: Ah, should have read further down :smile:. So I think from an API usage perspective, having the graph generated in the class is probably preferable. That way, you can remove all the sanity checks as well.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        Args:
+            node1 (tuple[int,int]): node1
+            node2 (tuple[int,int]): node2
+
+        Returns:
+            bool: whether both are neighbors.
+        """
+        return (abs(node1[0] - node2[0]) == 1 and node1[1] == node2[1]) or (
+            abs(node1[1] - node2[1]) == 1 and node1[0] == node2[0]
+        )
```
pehamTom, line 136: Given the rigid structure of the graph, is there even a need to have your layout be a networkx graph? Would an array not do the job here? Not requesting a rewrite of everything, just curious.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+        Returns:
+            bool: whether both are neighbors.
+        """
+        return bool(
+            abs(node1[1] - node2[1]) == 1 and abs(node1[0] - node2[0]) == 1
+        )  # diagonal neighbors or no neighbors at all
+
+    def collect_qubit_positions(self) -> list[tuple[tuple[int, int], tuple[int, int]]]:
+        """Collect the edges defining the qubits we need, depending on rough/smooth edges.
```
pehamTom: What does "we need" mean in this case? Is there a predicate that could be stated in the docstring?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                # common_el = sum(1 for t in star_set if t in rough_b)
+                # check whether the central node of the star is on the rough boundary
+                all_nodes = [node for edge in star for node in edge]
+                node_counts = Counter(all_nodes)
+                # central_node = max(node_counts, key=node_counts.get)
+                central_node = max(node_counts, key=lambda x: node_counts.get(x, 0))
+                common_el = central_node in rough_b
+                # print("common el", common_el)
+                lst_on_rough.append(common_el)
+            # print("lst_on_rough", lst_on_rough)
```
pehamTom: Comments left in the code.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+            assert sum(1 for x in lst_on_rough if x != 0) <= 1, (
+                "The star has overlaps with both rough boundaries, this cannot be."
+            )
```
pehamTom, line 211: Good use of an assertion :+1:

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+            # horizontal rough remove: weigh3 star with 2 overlapping nodes on rough b
+            if (
+                sum(lst_on_rough) == 1 and len(star) == 2
+            ):  # (sum(lst_on_rough)==2 and len(star)==3) #the weight3 case is actually not necessary right
+                pass
+            else:
+                stars_new.append(star)
+
```
pehamTom:
```suggestion
            if  sum(lst_on_rough) != 1 or len(star) != 2:  
                stars_new.append(star)

```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        self.stars = stars_new
+        return stars_new
+
+    def gen_plaquettes(self) -> list[list[tuple[tuple[int, int], tuple[int, int]]]]:
+        """Generates Plaquette Operators.
+
+        Returns:
+            list[list[tuple[tuple[int,int], tuple[int,int]]]]: _description_
```
pehamTom: Missing return value description?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                continue
+            collected.append(point)
+
+            j += 1
+            if j > max(t[1] for t in nodes):
+                j = 0
+                i += 1
+            if i > max(t[0] for t in nodes):
+                break
+        nodes += collected
```
pehamTom: Computation of `max` can be pulled out of the loop.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        """
+        plaquettes = []
+        nodes = self.boundary_nodes + self.inner_nodes  # all nodes
+        # since we check the nodes for being in the upper left corner, you can loose some plaquettes. hence add more nodes to minimum x and y, even though there will be useless checks
+        min_x = min(t[0] for t in nodes)
+        min_y = min(t[1] for t in nodes)
+        collected = []
+        i, j = 0, 0
+
+        while True:
```
pehamTom, line 235: I am not sure what happens in this loop. You start from a corner and then add nodes that are not in `nodes`. But isn't that all of them?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        # for each square, check where qubits are and add to plaquettes (number of qubits can be smaller than 4)
+        for node in nodes:
+            # create square of nodes with node in upper left corner.
+            square = [
+                node,
+                (node[0], node[1] + 1),
+                (node[0] + 1, node[1] + 1),
+                (node[0] + 1, node[1]),
+            ]  # cyclically aligned
+            # edges_square = list(zip(square, square[1:] + [square[0]]))
```
pehamTom:
```suggestion
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                # only check for those lements which are NOT the central qubit
+                node_counts = Counter(plaquette_flat)
+                # print("node_counts", node_counts)
+                # central_node = max(node_counts, key=node_counts.get)
+                central_node = max(node_counts, key=lambda x: node_counts.get(x, 0))
+                # print("central_node", central_node)
+                plaquette_flat = [node for node in plaquette_flat if node != central_node]
+                # print("plaquette_flat", plaquette_flat)
+                plaquette_set = set(plaquette_flat)
+                lst_on_smooth = []
```
pehamTom:
```suggestion
        for plaquette in plaquettes:
            if len(plaquette) == 2:  # 2edges= 2 qubits
                plaquette_flat = [node for edge in plaquette for node in edge]
                # only check for those lements which are NOT the central qubit
                node_counts = Counter(plaquette_flat)
                central_node = max(node_counts, key=lambda x: node_counts.get(x, 0))
                plaquette_flat = [node for node in plaquette_flat if node != central_node]
                plaquette_set = set(plaquette_flat)
                lst_on_smooth = []
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                for smooth_b in self.positions_smooth:
+                    common_el = sum(1 for t in plaquette_set if t in smooth_b)
+                    lst_on_smooth.append(common_el)
+                assert sum(1 for x in lst_on_smooth if x != 0) <= 1, (
+                    "The plaquette has overlaps with both rough boundaries, this cannot be."
+                )
+                if sum(lst_on_smooth) == 2:
+                    pass
+                else:
+                    plaquettes_new.append(plaquette)
```
pehamTom:
```suggestion
                if sum(lst_on_smooth) != 2:
                    plaquettes_new.append(plaquette)
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+        self.plaquettes = plaquettes_new
+        return plaquettes_new
+
+    def create_stabs(
+        self,
+    ) -> tuple[
+        list[list[tuple[tuple[int, int], tuple[int, int]]]], list[list[tuple[tuple[int, int], tuple[int, int]]]]
+    ]:
+        """Summarizes all Methods here."""
```
pehamTom: Is this a placeholder docstring?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        _ = self.fill_snake()
+        _ = self.collect_qubit_positions()
+        _ = self.gen_stars()
+        _ = self.gen_plaquettes()
+
+        q = len(self.qubit_edges)
+        lenstars = len(self.stars)
+        lenplaq = len(self.plaquettes)
+        # assert q - lenstars - lenplaq == 1, f"Your stabilizers are wrong. They should create 1 logical qubit but they yield {q - lenstars - lenplaq} instead."
+        # print("logical qubits", q - lenstars - lenplaq)
```
pehamTom:
```suggestion
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        if q - lenstars - lenplaq != 1:
+            warnings.warn(
+                f"Your stabilizers are wrong. They should create 1 logical qubit but they yield {q - lenstars - lenplaq} instead.",
+                UserWarning,
+                stacklevel=2,
+            )
```
pehamTom, line 320: Sanity check: Can this ever happen?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+        return self.plaquettes, self.stars
+
+        # add check whether there are q-num stabs = 1
```
pehamTom: I think you already did that check, right?

```suggestion
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+            if len(plaquette) == 6:  # pairs, weight-3, weight-5 in the interface NOT wanted for X stabs
+                # also remove the hex plaquettes within the interface (touching vertices of two triangles)
+                bools = []
+                for i in range(len(self.positions) - 1):
+                    if set(plaquette) & set(self.positions[i]) and set(plaquette) & set(self.positions[i + 1]):
+                        bools.append(False)
+                    else:
+                        bools.append(True)
+                if all(bools):
+                    x_plaquettes.append(list(plaquette))  # only if above NOT fulfilled
```
pehamTom: No need to use a list of bools here. 
```suggestion
                is_x_plaquette = True
                for i in range(len(self.positions) - 1):
                    if set(plaquette) & set(self.positions[i]) and set(plaquette) & set(self.positions[i + 1]):
                        is_x_plaquette = False
                        break
                if is_x_plaquette:
                    x_plaquettes.append(list(plaquette))  # only if above NOT fulfilled
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+            elif len(plaquette) == 4:
+                x_plaquettes.append(list(plaquette))
+
+        self.total_nodes = total_nodes
+        # build in assertion regarding number of each stabilizers, i have equations to check whether the number is right.
```
pehamTom: The sentence here seems a bit broken?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+        plt.show()
+        # !todo store also the pdf of the figure in a given path.
+
+    def gen_check_matrix(self, plaquettes: list[list[tuple[int, int]]]) -> npt.NDArray[np.int_]:
+        """Takes plaquettes and translates with self.integer_labeling."""
+        self.integer_labeling()
+        h = np.zeros((len(plaquettes), self.q_tilde), dtype=int)
+        for row, plaquette in enumerate(plaquettes):
+            if self.trans_dict is not None:  # case distinction only for mypy
```
pehamTom: You can add an `assert self.trans_dict` as well. Should even work if you do it after `self.integer_labeling` above. Mypy should be able to infer the type this way as well.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                translated_plaquette = [int(self.trans_dict[node]) for node in plaquette]
+            for el in translated_plaquette:
+                h[row, el] = 1
```
pehamTom: Use numpy indexing instead of a loop.
```suggestion
            h[row, translated_plaquette] = 1
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                node for plaq in next_ancilla_pairs for node in plaq
+            ]  # all nodes within the adjacent stdw
+            # find those nodes in the triangle boundary which are next to stdw and complement
+            inner_nodes = []
+            for node in nodes_stdw:
+                neighbors = list(self.g.neighbors(node))
+                boundary_neighbors = [n for n in neighbors if n in all_boundary]
+                inner_nodes += boundary_neighbors
+            inner_nodes = list(set(inner_nodes))  # remove duplicates
+            outer_nodes = [n for n in all_boundary if n not in inner_nodes]
```
pehamTom, line 756: Can these be sets?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        filtered_z_plaquettes += [lst_corner, lst_boundary]
+        [lst_corner, lst_boundary] = self.find_triangle_edges_corners(triangle_idx - 1)
+        filtered_z_plaquettes += [lst_corner, lst_boundary]
+
+        triangle_idx -= 1
+        for tup in itertools.combinations(self.outer_nodes_total[triangle_idx], 2):
+            for plaquette in filtered_z_plaquettes:
+                if tup[0] in plaquette and tup[1] in plaquette:
+                    subset_stabs.append(plaquette)
+                    continue
```
pehamTom: I think the `continue` doesn't do anything here since the loop body is over after this line.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                flag_filler = False
+
+        triangle_idx += 1
+        subset_stabs_temp = []
+        for plaq in subset_stabs:
+            overlap_left = sum(1 for node in plaq if node in self.positions[triangle_idx - 1])
+            overlap_right = sum(1 for node in plaq if node in self.positions[triangle_idx + 1])
+            if overlap_left > 2 or overlap_right > 2:
+                continue
+            subset_stabs_temp.append(plaq)
```
pehamTom: Given that the loop ends right after anyway, you could also do 
```suggestion
            if overlap_left <= 2 and overlap_right <= 2:
                subset_stabs_temp.append(plaq)
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+        return subset_stabs_temp
+
+    # helper functions for fill_triangle
+    @staticmethod
+    def get_single_nodes(subset_stabs: list[list[tuple[int, int]]]) -> list[tuple[int, int]]:
```
pehamTom: If this is only a private function not intended to be used via the public class interface you can write `_get_single_nodes` to indicate that it is private. As always in Python, the method is still accessible but it is convention.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+    def find_stabilizers_zz(self) -> list[list[tuple[int, int]]]:
+        """Summarizes the methods above and joins the subsets per triangle on the STDW."""
+        n = len(self.positions)
+        subset_z_stabs = []
+        self.find_outer_bdry()
+        for triangle_idx in range(1, n - 1):
+            subset_stabs = self.fill_triangle(triangle_idx)
+            subset_z_stabs += subset_stabs
+        # subset_z_stabs = list(map(set, {frozenset(s) for s in subset_z_stabs}))
```
pehamTom:
```suggestion
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        for plaq in subset_z_stabs:
+            overlap_left_logical = 0
+            overlap_right_logical = 0
+            for node in plaq:
+                if node in left_logical:
+                    overlap_left_logical += 1
+                if node in right_logical:
+                    overlap_right_logical += 1
+            if overlap_left_logical > 2 or overlap_right_logical > 2:
+                # subset_z_stabs.remove(plaq)
```
pehamTom:
```suggestion
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                    if neigh in interface_ancillas_flat:
+                        pair = [node, neigh]
+                if pair is None:
+                    msg = "the zlzl operator cannot be constructed with the given stabilizers"
+                    raise RuntimeError(msg)
+                # add pair of nodes to subset_z_stabs
+                subset_z_stabs.append(pair)
+
+        # final removal of duplicates
+        # return list(map(set, {frozenset(s) for s in subset_z_stabs}))
```
pehamTom:
```suggestion
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        return [list(set(s)) for s in {frozenset(s) for s in subset_z_stabs}]
+
+    def test_zz_stabs(self, subset_z_stabs: list[list[tuple[int, int]]]) -> bool:
```
pehamTom: Could this logic on the graph be replaced by checking whether all columns in `gen_check_matrix` contain an even number of 1's?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+        flattened_nodes = [item for sublist in subset_z_stabs for item in sublist]
+        counts = Counter(flattened_nodes)
+        for node, count in counts.items():
+            if count % 2 != 0 and node not in z_left_nodes and node not in z_right_nodes:
+                return False
+        return True
+
+
+class SnakeBuilder:
```
pehamTom: This name sort of implies that it generalizes the `SnakeBuilderSC` and `SnakeBuilderSTDW` classes. But this is not the case right? Would it be more appropriate to say `SnakeBuilderHex`?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+    def standard_steane() -> list[list[int]]:
+        """Returns the standard separate steane stabilizer plaquettes."""
+        return [[0, 2, 4, 6], [1, 2, 5, 6], [3, 4, 5, 6]]
+
+    def neighboring_vertex(self, vertex_0: tuple[int, int], vertex_1: tuple[int, int]) -> bool:
+        """Checks whether two vertices are adjacent."""
+        neighbor = False
+        if (vertex_0, vertex_1) in self.g.edges() or (vertex_1, vertex_0) in self.g.edges():
+            neighbor = True
+        return neighbor
```
pehamTom: Happens to the best from time to time, but you can just return the condition of the if-statement here.
```suggestion
    def neighboring_vertex(self, vertex_0: tuple[int, int], vertex_1: tuple[int, int]) -> bool:
        """Checks whether two vertices are adjacent."""
        return (vertex_0, vertex_1) in self.g.edges() or (vertex_1, vertex_0) in self.g.edges()
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        # neighboring_two is a bool which determines whether the weight 8 stab would connect neighboring patches
+        for key_i in pos_i_new:
+            for key_i1 in pos_i1_new:
+                # Check if the pair of keys are neighbors using the neighboring_vertex function
+                if self.neighboring_vertex(key_i, key_i1):
+                    neighboring_pairs.append((key_i, key_i1))  # Store the pair
+                    # If we already found 2 pairs, return true
+                    if len(neighboring_pairs) == 2:
+                        neighboring_two = True
+        return neighboring_two
```
pehamTom:
```suggestion
        neighboring_pairs = []
        # neighboring_two is a bool which determines whether the weight 8 stab would connect neighboring patches
        for key_i in pos_i_new:
            for key_i1 in pos_i1_new:
                # Check if the pair of keys are neighbors using the neighboring_vertex function
                if self.neighboring_vertex(key_i, key_i1):
                    neighboring_pairs.append((key_i, key_i1))  # Store the pair
                    # If we already found 2 pairs, return true
                    if len(neighboring_pairs) == 2:
                        return True
        return False
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+    def generate_x_stabilizers(self) -> list[dict[tuple[int, int], int]]:
+        """Subsequently builds the X stabilizers. Focus on the big `trillerpfeifen` weight-8 stabilizers."""
+        x_stabilizers: list[dict[tuple[int, int], int]] = []
+        compatible_x_stabs = self.compatible_x_stabs()
+
+        # check at which interface the next steane tile is placed
+        adjacent_edge = self.check_interface(0)
+
+        if sorted(adjacent_edge.values()) == sorted([0, 4, 3]):
```
pehamTom: Is there a reason not to just have `[0, 3, 4]`?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                key: value for key, value in self.positions[1].items() if value in compatible_x_stabs[k]["i+1"]
+            })
+            x_stabilizers.extend((
+                x_stab,
+                {key: value for key, value in self.positions[0].items() if value in {1, 2, 5, 6}},
+                {key: value for key, value in self.positions[0].items() if value in {3, 4, 5, 6}},
+                {key: value for key, value in self.positions[1].items() if value in {1, 2, 5, 6}},
+                {key: value for key, value in self.positions[1].items() if value in {3, 4, 5, 6}},
+            ))
+        elif sorted(adjacent_edge.values()) == sorted([1, 5, 3]):
```
pehamTom: Same as above?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        else:
+            msg = "Wrong edge connected between Steane patches."
+            raise RuntimeError(msg)
+
+        standard_steane_plaquettes = self.standard_steane()
+        # remaining steane patches
+        for i in range(1, len(self.positions) - 1):
+            current_patch = self.positions[i]
+            next_patch = self.positions[i + 1]
+            # adjacent_edge = self.check_interface(i)
```
pehamTom:
```suggestion
```

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+
+    @staticmethod
+    def compatible_x_stabs() -> list[dict[str, list[int]]]:
+        """Returns the allowed Weight 8 X Stabilizers crossing Steane patches."""
+        return [
+            {"i": [0, 2, 4, 6], "i+1": [0, 2, 4, 6]},
+            {"i": [3, 4, 5, 6], "i+1": [1, 2, 5, 6]},
+            {"i": [3, 4, 5, 6], "i+1": [0, 2, 4, 6]},
+            {"i": [1, 2, 5, 6], "i+1": [1, 2, 5, 6]},
+        ]
```
pehamTom, line 1039: Does this have to be a list of dicts instead of a list of tuples?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+            {"i": 1, "i+1": 1},
+            {"i": 0, "i+1": 0},
+            {"i": 3, "i+1": 1},
+            {"i": 1, "i+1": 3},
+            {"i": 3, "i+1": 0},
+            {"i": 0, "i+1": 3},
+            {"i": 1, "i+1": 0},
+            {"i": 0, "i+1": 1},
+        ]
+
```
pehamTom, line 1054: Same as above.

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+        for dictionary in dict_list:
+            # Get all integers in the dictionary's values
+            value_set = {val for values in dictionary.values() for val in values}
+            # Check if the two sets are disjoint
+            if pair_set.isdisjoint(value_set) and neighboring_pair(current_patch, next_patch, dictionary):
+                return dictionary
+
+        return None
+
+    def translate_checks(self) -> tuple[list[list[int]], list[list[int]]]:
```
pehamTom, line 1317: There is some inconsistency here. We have `gen_checks`, `gen_check_matrix` and `translate_checks` in the different classes, respectively. 

Also is there a reason these are lists instead of numpy arrays?

src/mqt/qecc/co3/microscopic/snake_builder.py
```diff
@@ -0,0 +1,1389 @@
...
+                # Order vertices and compute the convex hull for other faces
+                face_coords = convex_hull([pos[label] for label in face_vertices])
+                polygon = Polygon(face_coords, closed=True, edgecolor="black", facecolor=color, alpha=0.7)
+                plt.gca().add_patch(polygon)
+
+        plt.gca().set_aspect("equal")  # Ensure the aspect ratio is equal for proper visualization
+        plt.show()
+
+
+def convex_hull(points: list[tuple[float, float]]) -> list[tuple[float, float]]:
```
pehamTom, line 1373: Out of curiosity: Is this a version of the Graham Scan :smile: ?

pehamTom (changes requested):
Ok, so I finished my review until `snake_builder.py`. Since this is quite a large contribution I would give you the first chunk already and look at the rest later. If some of my questions are resolved by looking at the rest of the code then just ignore them.

The code looks good to me, I just added some suggestions, comments and questions where I thought some things could be improved.

docs/Co3.rst
LSHerzog: In principle yes. The rst file is just used as intermediate solution as long as this branch is not merged into main. However, the main branch's readme currently links to this rst file, which is why I would like to keep it for now. I would delete it as soon as we merge my branch into main.

docs/co3.md
LSHerzog: It should not be too hard, we only could not do it due to the deadline. I add some sentences to the md referencing parts of the paper where it is described. It would mainly require to adapt the Dijkstra algorithm to consider only those valid paths.

docs/co3.md
LSHerzog: Makes total sense, will adapt this
I also add links to the files in `scripts`, however they will work as soon as the branch is merged to main.

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: I am renaming tuple[int,int] as NodePos. This makes the type descriptions maybe a little clearer, but I have a lot of different nested types, so I would not define all of them separately.

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: I thought about this as well, but one has to define the `positions_rough` and `positions_smooth` beforehand following the numbering of the graph. Thus, a user needs to create the graph beforehand anyways, in order to know the potential labels. Because of this, I guess it should be okay to use the graph as input as well.

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: I used some in-built functions from networkx, which made it more convenient to use a networkx graph. But I guess one could also get rid of this dependency if one really wants it^^

burgholzer: Just a small side comment because I stumbled upon this in my email inbox: it might be worth checking out `rustworkx` as an alternative to `networkx`.
It's written in Rust instead of Python and provides a lot of the same functionality.
It also directly comes as a dependency of Qiskit, so it doesn't even add that much.
Not a must-have of course. Just something that might be worth considering. 🙂

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: I'm not 100% sure anymore as well, but I remember that before I added this, the code missed some plaquettes. The code checks for plaquettes starting in the upper left corner, but it may be possible that the upper left corner is not within the boundaries of the merged snake object (for plaquettes along diagonal boundaries), and then the stabilizer would be forgotten. Thus more nodes are added such that no stabilizers are missed. Not sure whether this helps^^

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: At least it happened to me if I used inputs which were not perfectly right. I rather have one warning too many than missing one^^

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: the comment is removed, this was just a todo comment for myself

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: You mean that i would avoid `inner_nodes` and `outer_nodes` being lists such that they are sets all the time? I think in principle yes, because the order does not matter here in principle. but I can imagine that changing the type may also affect the types for other functions and I'd rather like to avoid having new chaos with mypy^^

src/mqt/qecc/co3/microscopic/snake_builder.py
LSHerzog: Yes this was the very first idea only for the Steane Code and without STDWs. I could have deleted it as well, but I think it also does not hurt to keep it. I renamed it to `SnakeBuilderSteane`.

src/mqt/qecc/co3/microscopic/snake_builder.py
pehamTom: For me, this was not really about the dependency, but whether it would make the overall implementation simpler if one didn't use a graph here. Ultimately, it is not really an issue since you have to use graphs for other connectivities anyway.

src/mqt/qecc/co3/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+"""Co3."""
+
+from __future__ import annotations
+
+# from . import plots
```
pehamTom, line 12:
```suggestion
```

docs/Co3.rst
pehamTom: Then I'll leave this comment open until that time so we don't forget.

PR #390, opened 2025-04-03
Remove C++ part of package.

pehamTom:
## Description

This PR removes the C++ part of the package since it is no longer actively developed, and its features have largely been replaced by other libraries.

In particular, this removes the following features:

- C++ implementation of the union-find QLDPC decoder. [ldpc](https://github.com/quantumgizmos/ldpc) is actively maintained and developed for a dedicated package.
- Applying error correction to quantum circuits. This functionality is mostly deprecated, not actively developed and no future work is planned.

This also removes the necessary CI. 

This PR also updates the documentation to use `myst` and builds the Python documentation using [sphinx autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#module-sphinx.ext.autodoc).

## Checklist:


burgholzer (changes requested):
Thanks for starting this effort!
I just had some time and wanted to note down some must have's for this breaking change. I'll probably expand this comment in a couple of edits (just on my phone right now)

A couple of things to remember:
- the CD pipeline will require changes because QECC becomes a pure Python package now. Look at the MQT Bench repo for inspiration
- the Python stub files need to be removed
- the clang-tidy and the cmake-format config needs to be removed
- some of the pre-commit checks can be removed. Most notably clang-format and CMake-format
- please add a major label to this PR as this marks a significant breaking change.
- a big fat note should be included in the documentation that states which functionality has been removed (not deprecated! Removed) and that people who rely on that functionality should check out a version of QECC with a lower major version.
- the code-ql workflow should be adapted to only run the python checks. See mqt-bench.

Edit1: 
- Given how this is now a pure Python project, scikit-build-core is no longer the most suitable build system. The preferred solution is hatch(ling) 🐣 including hatch-vcs for dynamically managing the versions.
- the cibuildwheel config can be removed
- the pyproject.toml should be cleaned of any remainders of bound C++ code (including build time dependencies and entry point scripts)

pehamTom:
Thanks for pointing these things out. Most of that was on my list anyway but I wanted to start the PR already to spot early what things get broken by certain commits. Will get back to you once it is ready.

burgholzer:
> Thanks for pointing these things out. Most of that was on my list anyway but I wanted to start the PR already to spot early what things get broken by certain commits. Will get back to you once it is ready.

Like that!
It's just so easy to forget things. Wanted to make sure that I at least scribbled down my thoughts 💭 
I'll wait for your review request now 😉

burgholzer:
Thanks for pushing this @pehamTom.
I just went through the whole repository and fixed or modernized a couple of things.

I had to leave now. So this is still incomplete. I'll come back to this later today or tomorrow.

The dependencies in this project are a true pain. It would really be great to either take over maintenance of some of these projects, fix their build systems, or vendor the necessary code as part of QECC. The current situation is very much unsatisfactory.

One thing that I won't have time to fix is the documentation. The build reports several warnings of things not being found and some of the code examples are actually broken. Would be great if you could fix those.

pehamTom:
> > Thank you very much for that. I fixed most of the references that were not found in the documentation. The type aliases are not found but I think we will just have to live with that for now. Regarding the broken python code, I only found one example, have I overlooked anything?
> > When it comes to the dependencies, what issues do you see?
> 
> Many thanks. Just looked at the RtD build. I think the API docs look good now. Any further errors there can be ignored for now. However:
> 
>     * https://mqt--390.org.readthedocs.build/projects/qecc/en/390/LightsOutDecoder.html is broken with errors in the code cells and images not displaying correctly
> 
>     * https://mqt--390.org.readthedocs.build/projects/qecc/en/390/StatePrep.html also has broken static images
> 
>     * https://mqt--390.org.readthedocs.build/projects/qecc/en/390/StatePrep.html#circuits-and-evaluations has the wrong heading level and is listed as a top-level entry (`#` instead of `##` or lower)
> 
>     * https://mqt--390.org.readthedocs.build/projects/qecc/en/390/Encoders.html# should be using the matplotlib drawer. Otherwise the latex build of the docs will surely fail (you can try locally with `nox -s docs -- -b latex && cd docs/_build/latex && make`)
> 
>     * https://mqt--390.org.readthedocs.build/projects/qecc/en/390/Publications.html is missing publications that have been added to the Readme (and potentially others, haven't checked). These should also be cross referenced in the respective example notebooks to increase visibility.
> 
> 
> With regard to the dependencies: After the last couple of commits from yesterday, everything builds at least. However, there are several things that are less than ideal:
> 
>     * the `qecsim` package (which is hardly maintained) is only used for a paper comparison. Yet it is included as a full dependency.
> 
>     * there is still an upper cap on `ldpc` because QECC is not compatible yet.
> 
>     * we include the `bposd` package (which is hardly maintained) although similar functionality seems to be available in `ldpc`
> 
>     * we include the `qsample` package (which is not maintained and poorly configured)
> 
> 
> Obviously, these are not things to be solved as part of this PR. But it would really be great to clean this up in the near term future.

Thanks for pointing these errors out. The images went under my radar. The automatic conversion to markdown didn't work properly here. Except for the weird sdist failure in the CI, all should be done now as far as I see.

burgholzer (approved):
Thanks for all your work on this. 
I just briefly skimmed everything one last time after pushing some finishing touches. (also fixed the latex build, which, as I expected, was just latex's incapability of handling unicode again)
This is looking pretty smooth now and is a big step towards v2.
Before releasing that, I'd tie up some loose ends and, most importantly, move the repository to the new organization and deal with everything that comes with that.

Once CI is green. This is good to go though.

PR #420, opened 2025-06-16
Fix setuptools version problem.

pehamTom:
## Description

Currently, the CI is failing everywhere because of a deprecated API used by the qsample dependency. This PR fixes this issue.

## Checklist:

- [ ] The pull request only contains commits that are related to it.
- [ ] I have added appropriate tests and documentation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ dependencies = [
   "qsample>=0.0.2",
   "urllib3>=2.3.0", # Required by qsample (to be removed)
   "fastcore>=1.8.1", # Required by qsample (to be removed)
+  "setuptools<81", # Version cap required by qsample (to be removed)
 ]
 dynamic = ["version"]
 
@@ -230,6 +231,7 @@ wille = "wille"
 ser = "ser"
 aer = "aer"
 anc = "anc"
+arange = "arange"
 
 
 [tool.repo-review]


pehamTom:
> Random side comment here: we should really resolve this messy dependency at some point. I still remember that we already discussed options for this a while back:
> 
>     * vendor the relevant code from qsample into qecc and remove the dependency
> 
>     * fix the packaging in the qsample repository
> 
>     * fork the qsample repo and update the packaging

I would like to avoid a situation where we maintain another repository. My initial fix that I wanted to make was to make qsample optional, as it is only used in the simulation of the deterministic circuits. But since the simulation functionality is actively tested, this wouldn't remove errors in the CI.

 If version capping setuptools doesn't introduce further problems right now, I would fix the issue like this in this PR. If (or rather, as soon as) this leads to further problems, I would then make the simulation optional or deprecate that functionality.

pehamTom:
> This looks like a reasonable fix. It's unfortunate that the respective functionality is no longer tested, but as you said in the PR, that's the only way to avoid the CI errors without spending more time on warning filters, etc.
> 
> You are missing one crucial step for this PR to be ready to merge and that is an update to the lockfile. This is needed to reflect the updated dependency structure. A simple `uv lock` should be sufficient here. I'd probably do `uv lock -U` to also directly update the versions in the lockfile and require one less PR to go through CI.

Thanks for looking over this and pointing out the lockfile. Even had the updated lockfile but reverted again to try something. It is online now.

burgholzer (approved):
Nice 👍🏻 should be good to go now!

PR #426, opened 2025-06-25
Ignore qsample warnings.

pehamTom:
## Description

Recently, qsample has been made an optional dependency, which means that tests relying on qsample are skipped in the CI and don't lead to failures due to DeprecationWarnings anymore. However, if qsample is installed locally and tests are run, the tests will still fail. 

This PR fixes this issue by ignoring the warnings thrown by qsample.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,6 +96,8 @@ log_cli_level = "INFO"
 xfail_strict = true
 filterwarnings = [
   "error",
+  "ignore:pkg_resources is deprecated:UserWarning", # Due to qsample, do not remove!
+  "ignore:Deprecated call to `pkg_resources\\.declare_namespace\\('.+'\\)`\\.:DeprecationWarning", # Due to qsample, do not remove!
   "ignore:.*The retworkx package is deprecated*:DeprecationWarning:pymatching",
   'ignore::DeprecationWarning:.*(simpleeval).*',
   # Qiskit 1.3 deprecations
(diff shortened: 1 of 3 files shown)


burgholzer (changes requested):
Actually, I would very much prefer to simply add the corresponding filterwarnings to the `pyproject.toml`.
That is a totally acceptable solution and considerably simplifies the diff here. The solution here is convoluted and I am not 100% sure that it even works because you still have a line like 
```
from qsample import Circuit
```
Which will still run the code from `qsample.__init__.py` that would probably trigger the deprecation warnings.

The filter warnings are really not a "taboo" thing. Their purpose is exactly to ignore warnings that one has no control over. Such warnings should not overcomplicate the code of the library.

burgholzer:
> My reasoning was to only filter the warning if it comes from qsample. If (for some reason) another dependency is non-optional dependency causes issues with the same warning, we might want to catch that. It is an unlikely possibility though, I agree. Gonna change it then.
> 

Filter warnings actually have a dedicated option to do exactly that. One of the arguments in the warning filter specification can be used to filter only warnings from a particular origin.
From experience, this is fairly finicky to set up and get right, which is why we go with the unconditional ignore most of the time.
You could still try that here though.
The details can be found here: https://docs.python.org/3/library/warnings.html#the-warnings-filter

I'll approve this PR for now. If you still want to try to make the warning filters more specific, feel free to go for it 😌

> 
> Regarding the code. Hard to see in the CI but since the import only happens in the if statement for type checking, the code is not run during testing (at least on my machine). 

Ah. That makes sense. I indeed overlooked this in the git diff. 👍🏻

burgholzer (approved):

PR #437, opened 2025-07-08
Fix minor typo in `CatStates.md`

vtomole:


pehamTom:
Thanks for spotting and fixing this!

PR #440, opened 2025-07-09
👷 Enable templating

denialhaag:
## Description

This PR adds a workflow that synchronizes common files with rendered templates.

The templates can be found in the https://github.com/munich-quantum-toolkit/templates repository.

## Checklist:

- [x] ~~I have added appropriate tests and documentation.~~

### DIFF
--- a/.github/workflows/templating.yml
+++ b/.github/workflows/templating.yml
@@ -1,0 +1,25 @@
+name: Templating
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/templating.yml"
+  workflow_dispatch: # Allow manual triggering
+
+jobs:
+  render-template:
+    name: Render template
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


denialhaag:
> > > @burgholzer, before merging this PR, we should probably add a changelog and an upgrade to the repository, right? Otherwise, the updated PR template doesn't make a lot of sense. 🤔
> > 
> > 
> > I remember some discussion around this with pehamTom but I don't recall what the conclusion was. We could also (temporarily) disable the synchronization for the PR template here.
> 
> Okay, just talked with Tom, and apparently I am hallucinating more than LLMs these days. Adding a changelog and upgrade guide here should be totally fine. 👍🏼 best done in a separate PR, though, I suppose.

The changelog and the upgrade guide are added with #442. 👍

burgholzer (approved):
LGTM 👍🏼

PR #662, opened 2026-04-03
Replace unreachable early return with assert in _check_code_correct

maayanmatsliah-tech:
Motivation: The second early return in _check_code_correct is unreachable — the block above already raises InvalidStabilizerCodeError if exactly one logical is None, so by the time execution reaches the second check, both are guaranteed to be None or neither is. Replacing it with an assert makes this explicit and allows mypy to narrow the type correctly for the rest of the method.
Testing: No behavior change. Existing tests cover this.
Documentation: No documentation updates needed.


burgholzer (approved):
Thanks

PR #691, opened 2026-05-15
🔧 Replace `mypy` with `ty`

denialhaag:
## Description

This PR replaces `mypy` with `ty` and fixes all found errors.

## Checklist

- [x] ~~I have added appropriate tests that cover the new/changed functionality.~~
- [x] ~~I have updated the documentation to reflect these changes.~~


src/mqt/qecc/circuit_synthesis/cat_states.py
```diff
@@ -415,14 +417,14 @@ def fault_gens_from_circuit(circ: stim.Circuit) -> list[int]:
...
     circ = stim.Circuit()
-    circ.append("RX", [4])
-    circ.append("R", [0, 1, 2, 3])
-    circ.append("CX", [4, 0])
-    circ.append("CX", [0, 1])
-    circ.append("CX", [1, 2])
-    circ.append("CX", [2, 3])
-    circ.append("CX", [3, 4])
-    circ.append("MR", [4])
+    circ.append("RX", [4])  # ty: ignore[no-matching-overload]
```
burgholzer, line 420: Is there really no better way to add these gates to the circuit that does not trigger these kinds of warnings?

src/mqt/qecc/circuit_synthesis/elimination.py
```diff
@@ -383,6 +385,8 @@ def _maybe_use_best_solution(
     if not hasattr(generator, "get_best_solution") or not hasattr(generator, "score_fn"):
         return current_ops, current_tableau
 
+    generator = cast("RolloutCandidateGenerator", generator)
+
```
burgholzer, line 389: If this cast is necessary, shouldn't the function signature be updated instead?

src/mqt/qecc/circuit_synthesis/strategy.py
```diff
@@ -68,8 +68,9 @@ def for_cnot_up_to_row_ops(
     cached_rank: int | None = None
 
     def termination_criterion(tbl: BinaryMatrix) -> bool:
+        assert isinstance(tbl, CheckMatrix)
```
burgholzer: shouldn't the signature of the function be updated instead?

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
```diff
@@ -54,14 +54,14 @@ def run_with_timeout(func: Callable[[Any], Any], *args: Any, timeout: int = 10)
...
-    fun: Callable[[int], QuantumCircuit],
+    fun: Callable[[Any], Any],
     min_param: int,
     max_param: int,
     min_timeout: int,
     max_timeout: int,
     param_factor: float = 2,
     timeout_factor: float = 2,
-) -> tuple[QuantumCircuit | None, int] | None:
+) -> tuple[Any, int] | None:
```
burgholzer: why the loss in type information here?

pyproject.toml
```diff
@@ -306,4 +320,5 @@ test = [
 dev = [
   {include-group = "test"},
   "nox>=2025.11.12",
+  "ty==0.0.34",
```
burgholzer:
```suggestion
  "ty==0.0.36",
```
Given how there are a couple more changes requested as part of this review, we may as well switch to the latest version of `ty` here already.

burgholzer (changes requested):
Thanks @denialhaag for going through this chore! 🙏🏼 
This is looking great overall!
The part that is most annoying is the ignores for the stim circuit definitions. It would be really nice if there where a way to avoid having to place an ignore there every time.

src/mqt/qecc/circuit_synthesis/cat_states.py
denialhaag: I really think this is the correct usage: https://github.com/quantumlib/Stim/blob/e9cc7dee0a67bb4dbea2733f0170d03dcecd3cd0/doc/python_api_reference_vDev.md#stim.Circuit.append

I don't know if this is an issue of `ty` being too picky (?) in its current version. I'd say `[2, 3]` should fall into `Union[<...>, Iterable[Union[int, <...>]]]`. But this is definitely exactly the usage described in the docs linked above. 🤔

src/mqt/qecc/circuit_synthesis/elimination.py
denialhaag: In principle, the function could be called by any generator implementing the `CandidateGenerator` interface:

https://github.com/munich-quantum-toolkit/qecc/blob/f02df8e926cf1eaba7d37679e6634e0d08950b74/src/mqt/qecc/circuit_synthesis/elimination.py#L356-L361

The `EliminationStrategy` is defined to have a `CandidateGenerator` and not specifically a `RolloutCandidateGenerator`. Changing the typing in `RolloutCandidateGenerator` would have quite some ripple effects.

The way I understand this now is that the case above (`not hasattr(generator, "score_fn")`) captures a non-`RolloutCandidateGenerator` generator. That said, the code base currently does not have any implementation of a non-`RolloutCandidateGenerator` generator.

I'm not saying that the implementation couldn't be improved, but there are some nuances here.

src/mqt/qecc/circuit_synthesis/strategy.py
denialhaag: This is another consequence of the `EliminationStrategy` interface: The function is subsequently passed to its constructor, and this expects a function expecting a `BinaryMatrix` (which is defined as `CheckMatrix | StabilizerTableau`):

https://github.com/munich-quantum-toolkit/qecc/blob/f02df8e926cf1eaba7d37679e6634e0d08950b74/src/mqt/qecc/circuit_synthesis/strategy.py#L134-L139

I have slightly adapted the solution in 1a2bfa4d93f104eda2d93c3f0acdc7a0c060398d. This is more in line with how the function is defined in other places.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
denialhaag: I have slightly improved the annotation of the `fun` argument in https://github.com/munich-quantum-toolkit/qecc/pull/691/commits/37fcf0455675deae5d0503e71f61c2d2e3d43f8b. 

I cannot immediately find an instance where this function ever returned a `QuantumCircuit`. The function is called three different times. The signatures of the input `fun` are:

- `def _func(num_anc: int) -> Recovery | None:`
- `def fun(num_cnots: int) -> list[npt.NDArray[np.int8]] | None:`
- A `functools.partial` function whose return type I cannot untangle right now

pyproject.toml
denialhaag: Good point! Done locally because I also wanted to update the lock file directly. There were no new errors. 😌

src/mqt/qecc/circuit_synthesis/cat_states.py
denialhaag: I kind of "like" putting `ty: ignore` comments in such cases because `ty` will complain and ask to remove them once they are no longer necessary (in case an update of `stim` or `ty` fixes the underlying issue).

src/mqt/qecc/circuit_synthesis/elimination.py
burgholzer: Ah. I see. In that case, let's not make this too disruptive and stick with the current solution.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
burgholzer: Maybe @pehamTom can chime in here and clarify?

src/mqt/qecc/circuit_synthesis/cat_states.py
burgholzer: Hm, yeah. Feels like a ty limitation.
What is the actual error that Ty reports here?

denialhaag: It's referencing the same method signature:

```
error[no-matching-overload]: No overload of bound method `Circuit.append` matches arguments
     --> src/mqt/qecc/circuit_synthesis/cat_states.py:427:5
      |
  427 |     circ.append("MR", [4])
      |     ^^^^^^^^^^^^^^^^^^^^^^
      |
  info: First overload defined here
     --> .venv/lib/python3.13/site-packages/stim/__init__.pyi:300:5
      |
  300 | /     @overload
  301 | |     def append(
  302 | |         self,
  303 | |         name: str,
  304 | |         targets: Union[int, stim.GateTarget, stim.PauliString, Iterable[Union[int, stim.GateTarget, stim.PauliString]]],
  305 | |         arg: Union[float, Iterable[float]],
  306 | |         *,
  307 | |         tag: str = "",
  308 | |     ) -> None:
  309 | |         pass
      | |____________^ First overload defined here
      |
  info: Possible overloads for bound method `append`:
  info:   (self, name: str, targets: int | GateTarget | PauliString | Iterable[int | GateTarget | PauliString], arg: int | float | Iterable[int | float], *, tag: str = "") -> None
  info:   (self, name: CircuitInstruction | CircuitRepeatBlock | Circuit) -> None
  info: Overload implementation defined here
     --> .venv/lib/python3.13/site-packages/stim/__init__.pyi:316:9
      |
  316 |       def append(
      |  _________^
  317 | |         self,
  318 | |         name: object,
  319 | |         targets: object = (),
  320 | |         arg: object = None,
  321 | |         *,
  322 | |         tag: str = '',
  323 | |     ) -> None:
      | |_____________^
      |
```

burgholzer: Hm. It would probably be good to report this as a (potential) bug in the Ty repository. Should be a pretty simple reproducer.

denialhaag: Done: https://github.com/astral-sh/ty/issues/3480

burgholzer: Based on the comment there, this actually seems to be a bug in the stubs. If that turns out to be true, a PR to stim would be nice.

denialhaag: Yeah, I don't know how I missed that... The issue already seems to be fixed (compare the `ty` output with [the docs](https://github.com/quantumlib/Stim/blob/e9cc7dee0a67bb4dbea2733f0170d03dcecd3cd0/doc/python_api_reference_vDev.md#stim.Circuit.append)). We'll just have to wait for a release of `stim` then.

burgholzer: Ah, nice. At least something. In that case, we should be good here for now and I think this can be resolved.

src/mqt/qecc/circuit_synthesis/synthesis_utils.py
pehamTom: That seems to be a leftover from ancient times. This hasn't been used for functions returning `QuantumCircuit` in a while, I think. `Any` seems like the most sensible type here. The idea of `iterative_search_with_timeout` is to have a function taking some int parameter and repeatedly calling it until a solution is found (or it really times out).

We use this in several places in QECC for SAT-based searches where we iteratively increase the time budget. Empirically, it is better to quickly find the "easy" instance (based on the input parameter) and then try to refine the solution than to use the complete time budget for every search.

burgholzer (approved):
Nice work 🙌🏼

PR #734, opened 2026-07-02
Custom input states for heuristic State Preparation Circuit

MaxieHelenBichmann:
## Description

The construction of a heuristic state preparation circuit was extended by adding the possibility of choosing an arbitrary input state, which can be a string of 0 and + (previously: only a choice between all $|0\rangle_L$ and all $|+\rangle_L$ possible) for the given code.

Assisted-by: GitHub Copilot through inline suggestions; Codex via GPT-5.5 as first reviewer (both reviewed by me)

## Checklist

- [ ] The changes are fully tested and pass the CI checks.

### DIFF
--- a/docs/StatePrep.md
+++ b/docs/StatePrep.md
@@ -14,7 +14,7 @@ mystnb:
 # Fault tolerant state preparation of Pauli eigenstates for CSS codes
 
 The QECC package contains functionality for synthesizing and simulating fault tolerant and non-fault tolerant state preparation circuits for Pauli eigenstates of CSS codes.
-Currently it supports synthesizing circuits for preparing the $|0\rangle_L^k$ and $|+\rangle_L^k$ states of arbitrary $[[n,k,d]]$ CSS codes.
+Currently it supports synthesizing circuits for preparing the any logical state of the form $\{|s\rangle_L \mid s \in \{0,+\}^k\}$ of arbitrary $[[n,k,d]]$ CSS codes.
 
 ## Synthesizing non-FT state preparation circuits
 
--- a/docs/StatePrep.md
+++ b/docs/StatePrep.md
@@ -14,7 +14,7 @@ mystnb:
 # Fault tolerant state preparation of Pauli eigenstates for CSS codes
 
 The QECC package contains functionality for synthesizing and simulating fault tolerant and non-fault tolerant state preparation circuits for Pauli eigenstates of CSS codes.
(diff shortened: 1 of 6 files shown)


docs/StatePrep.md
```diff
@@ -14,7 +14,7 @@ mystnb:
 # Fault tolerant state preparation of Pauli eigenstates for CSS codes
 
 The QECC package contains functionality for synthesizing and simulating fault tolerant and non-fault tolerant state preparation circuits for Pauli eigenstates of CSS codes.
-Currently it supports synthesizing circuits for preparing the $|0\rangle_L^k$ and $|+\rangle_L^k$ states of arbitrary $[[n,k,d]]$ CSS codes.
+Currently it supports synthesizing circuits for preparing the any logical state of the form $\{|s\rangle_L \mid s \in \{0,+\}^k\}$ of arbitrary $[[n,k,d]]$ CSS codes.
```
pehamTom:
```suggestion
Currently, it supports synthesizing circuits for preparing any logical Pauli eigenstate of the form $\{|s\rangle_L \mid s \in \{0,+\}^k\}$ of arbitrary $[[n,k,d]]$ CSS codes.
```

src/mqt/qecc/circuit_synthesis/state_prep.py
```diff
@@ -200,7 +200,7 @@ def _build_state_prep_circuit_from_back(
 def heuristic_prep_circuit(
     code: CSSCode,
     optimize_depth: bool = True,
-    zero_state: bool = True,
+    zero_state: bool | str = True,
```
pehamTom: I am aware that I said that we should probably keep the `zero_state` flag for backward compatibility. But i think overloading that parameter probably leads to more confusion. I am also ok with breaking compatibility. We haven't had the new release yet (which breaks stuff anyway) so merging this before the new release should actually be fine. 

My suggestion: rename this to `state` and set it to "all_zero" by default. So in addition to strings of "0" and "+" we can allow "all_zero" and "all_plus" or something.

pehamTom (changes requested):
Generally looks good, but see my comment below for suggested (interface-breaking) changes to the function.

src/mqt/qecc/circuit_synthesis/state_prep.py
MaxieHelenBichmann: Yes, I was also unsatisfied with keeping the parameter with that naming and use, as it does not really fit the new semantics. I'll update it, thanks for the suggestion!

pehamTom (approved):
Thanks for the changes. LGTM!

PR #738, opened 2026-07-02
Support fixing input qubits in encoder circuit synthesis

MaxieHelenBichmann:
## Description

The synthesis of encoding circuits for `CSSCode` and `StabilizerCode` was extended to enable fixing one or multiple input qubits to either $|0\rangle$ or $|+\rangle$, effectively synthesizing a circuit for a code with a lower $k$.

_Note 1_: As the circuit effectively encodes a different code than the given input, functions like `logical_to_input_mapping` do not have the expected behavior in that case.
_Note 2_: As there might be a (unrelated) bug in `StabilizerCode.__init__` in the case of no logical operators in the form of empty lists, the case that $k$ input qubits are fixed is not tested yet. 

Assisted-by: GitHub Copilot through inline suggestions; Codex via GPT-5.5 as first reviewer (both reviewed by me)

## Checklist


src/mqt/qecc/circuit_synthesis/encoding.py
```diff
@@ -829,22 +829,41 @@ def synthesize_encoding_circuit(
     code: StabilizerCode,
     config: SynthesisConfig | None = None,
     use_cnots_if_css: bool = True,
+    fixed_logical_qubits: list[tuple[int, bool]] | list[int] | None = None,
```
pehamTom: I think it would be better to have a more descriptive parameter here than a Bool. Maybe a str that can be either "X" (|+>) or "Z" (|0>). Or simply "0" and "+" (your choice).
This of course then requires checking that the string is valid. Going further, one might consider introducing a specific type for these kinds of situations as this appears in multiple places in QECC. But this is not something to be handled in this PR. 

I also think that since this is a somewhat niche feature, it is fine to demand the user to be explicit. Hence, I would drop the `list[int]` version. 

Nit: Also, instead of a list of tuples, this can just be a `dict`, I think.

src/mqt/qecc/circuit_synthesis/encoding.py
```diff
@@ -829,22 +829,41 @@ def synthesize_encoding_circuit(
...
         code: The stabilizer code to synthesize the encoding circuit for.
         config: Configuration options for the synthesis process.
         use_cnots_if_css: Whether to use CNOT-only synthesis if the code is CSS.
+        fixed_logical_qubits: List of logical qubit indices that should be fixed to a logical zero state in the circuit, or list of tuples (index, is_zero) where is_zero indicates whether the logical qubit is fixed to a logical zero or plus state.
 
     Returns:
         A CliffordIsometry that implements the encoding circuit for the given stabilizer code.
     """
+    complete_fixed_logical_qubits = [
+        (item, True) if isinstance(item, int) else item for item in (fixed_logical_qubits or [])
```
pehamTom: Nit: the `isinstance` check only has to be done once. 

But this point is moot if the above comment is considered.

tests/circuit_synthesis/test_encoder_synthesis.py
```diff
@@ -405,3 +405,50 @@ def test_local_minimum() -> None:
     )
     enc = synthesize_encoding_circuit(code, config=config)
     assert enc.get_code().is_equivalent(code)
+
+
+@pytest.mark.parametrize("fixed_logical_qubits", [None, [], [0], [(0, True)], [(0, False)], [(1, True), (0, False)]])
```
pehamTom: To make sure we also catch edge-cases:

- Test non-contiguous blocks of logicals (i.e. [1,6,4]).
- Test invalid logical assignment (this should fail)

tests/circuit_synthesis/test_encoder_synthesis.py
```diff
@@ -405,3 +405,50 @@ def test_local_minimum() -> None:
...
+    """Check that fixed logical qubits are removed from the encoder inputs."""
+    encoder_cnots = synthesize_encoding_circuit(
+        css_4_2_2_code,
+        fixed_logical_qubits=fixed_logical_qubits,
+        use_cnots_if_css=True,
+    )
+    encoder_no_cnots = synthesize_encoding_circuit(
+        css_4_2_2_code,
+        fixed_logical_qubits=fixed_logical_qubits,
+        use_cnots_if_css=False,
```
pehamTom: Nit: handle setting of this parameter via another parameter in the test.

pehamTom (dismissed):
Thanks for your work here. This already looks pretty good. I have some rather minor suggestions for your consideration.

src/mqt/qecc/circuit_synthesis/encoding.py
MaxieHelenBichmann: Very valid point, anticipated something along those lines. Thanks for the feedback!

lsschmid:
@coderabbit review

PR #768, opened 2026-07-24
Phase-sensitive Algebra

### LINKED ISSUE #767: Restore phase-sensitive stabilizer equality/validity checks
Follow-up from #764.

To avoid invalid GF(2) phase handling, that PR moved the stabilizer-group equality and equivalence checks to the phaseless symplectic representation, so signs are currently ignored (see `equal_stabilizer_group`, `stabilizer_equivalent` in `stabilizer_code.py`).

Once we have sign-aware multiplication/rank for `PauliTableau`, we should:
- make `equal_stabilizer_group` / `stabilizer_equivalent` phase-sensitive again, and
- add the `-I ∉ S` validity check.

Related: decide whether `generators` in the `*Code` classes should be allowed to contain redundant Paulis (currently allowed).

MaxieHelenBichmann:
## Description

This PR extends the symplectic representation with a **phase-sensitive algebra** so that the additional phase-bits are transformed according to the properties of the Pauli matrices.

The Pauli-representation layer now supports phases of the whole **n-qubit  Pauli group** $\mathfrak{P}_n$, meaning a `Pauli` represents an element of the n-qubit Pauli group and the rows of a `PauliTableau` a subgroup of $\mathfrak{P}_n$. 
As the symplectic representation `( x | z )` removes phase information, `Pauli` contains the member `phase` (or property `phase_exponent`), which together define the n-qubit Pauli operator $i^{\text{phase}} X^{\text{x}} Z^{\text{z}}$. With this representation, Pauli multiplication and the relation $Y = iXZ$ can be fully simulated with the symplectic vector and the phase exponent.

I did *not* restrict the representation to **Hermitian** Paulis (in that case, `Pauli` would only have phases $\pm 1$) or to an Abelian subgroup without $-I$ (in that case, `PauliTableau` could use the commutation assumption to simplify phases to $\pm 1$ during Pauli-multiplication). In my *subjective* opinion, those restrictions should be enforced by the code layer above (`*Code`) and *not* the Pauli-representation. 
Consequently, however, `Pauli*` supports a slightly more complex algebra due to the possible phases $1, i, -1, -i$. I would also agree that there is no use case yet for the theoretically complete (but more complex) representation of the full Pauli group. So this is an open question.

However, even with a restricted version, the simple **Boolean addition of the symplectic vector** (concatenated with a phase bit) is not the correct algebraic operation to simulate phase-sensitive Pauli multiplication, since we would need to include a **correction term** for the phase bit that depends on the $X$/$Z$ overlap. Eg:
$P_1 = X \otimes X , P_2 = Z \otimes Z $ 
$P_1 * P_2 = (X \otimes X)*(Z \otimes Z) = (XZ) \otimes (XZ) = (-iY) \otimes (-iY) = i^2 (Y \otimes Y) = - Y \otimes Y$

Previous representation $(x | z | p)$ with $p \in \\{0,1\\}$ and $P = (-1)^p P_1 \otimes .. \otimes P_n$ and $(x_i | z_i) \sim I \text{ if } (0|0); X \text{ if } (1|0); Y \text{ if } (1|1); Z \text{ if } (0|1)$ meaning $P = (-1)^p i^{x * z} X^x Z^z$:
$P_1  \sim (00|11|0) , P_2  \sim (11|00|0)$ 
$(11|00|0) + (00|11|0) = (11|11|0) \sim (-1)^0 i^{2} (XX) (ZZ)  = Y \otimes Y \not = - Y \otimes Y$ thus normal binary addition $+$ does not work for the phases in this representation

Phase-sensitive calculation should be $(x | z | e)$ with $e \in \\{0,1, 2, 3\\}$ with $i^e X^x Z^z$:
$P_1 \sim (11|00|0) , P_2 \sim (11|00|0)$
$(11|00|0) \oplus (00|11|0) = (11|11|0) \sim  i^0 (XX) (ZZ) = -Y \otimes Y $
with $\oplus$ calculating the  new phase as $e = e_1 + e_2 + 2*(z_1 * x_2) \text{ mod } 4$

Thus, either way, the normal mod2 `row_echelon` cannot be used in the phase-sensitive context, and we would need a `pauli_ row_echelon` with the correction; nonetheless, no matter if we only restrict ourselves to Hermitian Paulis.

Another open question is whether there is a need to **document** the representation layers (or at least high-level classes like `*Code`) and their assumptions in the generated Website.

Additionally, we fix the assumption that `generators` of the code have to be **linearly independent**.

Fixes #767 

Assisted-by: GPT-5.6 Sol as first reviewer

## Checklist

- [x] Any agent that created, edited, or submitted GitHub content was explicitly authorized for that scope, as required by our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/qecc/blob/main/docs/ai_usage.md).
- [ ] Every agent-authored or agent-edited public text body begins with the visible disclosure `🤖 *AI text below* 🤖` (titles are exempt).


lsschmid:
Thanks @MaxieHelenBichmann, looks good so far.

I pushed two additional commits:

- **A few fixes.** Main one: `independent_rows()` in `StabilizerCode.__init__` also
  stripped redundant rows out of `CSSCode` (`Hx`/`Hz` are views into the generators), so
  check matrices came back shorter and syndromes with them. That independence was only
  needed for the `n_rows` comparison in `equal_stabilizer_group`, so I compare group
  order there instead and keep generators as given. Also: a crash when both codes have
  no stabilizers, `phase` → `phase_exponent` so old code fails loudly, `__getitem__` no
  longer aliasing, a `multiply_rows` helper, a vectorized `pauli_row_echelon` (n=761
  construction 2.4s → 0.8s, `is_stabilizer` 3.1s → <10ms), and a short docs page.
- **The SAT fix you predicted.** `_stabilizer_sign_corrections` XORs signs across a row
  reduction — for `YY`/`ZZ` the canonical generator is `YY * ZZ = -XX`, but the XOR says
  `+XX`. Swapped in your `pauli_row_echelon`. Non-CSS only, which is why nothing caught
  it; added a test that fails without the fix.

Looks fine now, going to merge once CI runns through

lsschmid:
@coderabbit review

lsschmid:
@coderabbitai review

PR #776, opened 2026-07-27
🐛 Normalize CSS check-matrix dtype to int8

### LINKED ISSUE #775: 🐛 CSSCode accepts uint8/int16 check matrices that later crash (or hang) state-preparation synthesis
### System and Environment Information

- Operating system and version: macOS 26.5.1 (arm64); also reproduced on the 1.9.0 release with a different symptom, see below
- MQT QECC version: 2.0.0 (PyPI) and main @ 0e208bf66 (2.0.1.dev14)
- Language versions: Python 3.13.13
- Any relevant dependencies and their versions: numpy 2.4.6, numba 0.66.0, ldpc 2.4.1, qiskit 2.5.1

### Bug Description

`CSSCode(distance, Hx, Hz)` accepts check matrices of any integer dtype (it validates orthogonality, but not dtype). `heuristic_prep_circuit` then behaves dtype-dependently: int8 and int64 matrices synthesize fine, while uint8 and int16 matrices crash with a numba dispatch error (`TypeError: No matching definition for argument type(s) array(int16, 2d, C), array(int64, 1d, C), ...`). uint8 is a natural dtype for GF(2) matrices, so this is easy to hit when constructing codes from your own parity checks.

On the previous release (1.9.0) the same uint8 inputs do not raise at all: synthesis keeps printing "Local minimum reached. Making matrix triangular." and had not returned after 60 s on the snippet below, while the int8 control returns immediately (on a 12-qubit code we counted over 7&times;10^5 restarts in under three minutes). That is the harder-to-diagnose variant of the same problem.

`CSSCode.from_code_name` itself returns int64 matrices while the constructor signature annotates int8, so the de-facto contract is wider than the annotation. Expected behavior: either normalize the dtype at construction, or reject unsupported dtypes there with a clear message.

### Steps to Reproduce

1. `pip install mqt.qecc==2.0.0`
2. Run:

```python
import numpy as np
from mqt.qecc import CSSCode
from mqt.qecc.circuit_synthesis import heuristic_prep_circuit

steane = np.array(
    [[1, 1, 1, 1, 0, 0, 0], [0, 1, 1, 0, 1, 1, 0], [0, 0, 1, 1, 0, 1, 1]]
)
code = CSSCode(distance=3, Hx=steane.astype(np.uint8), Hz=steane.astype(np.uint8))
heuristic_prep_circuit(code)  # TypeError from numba dispatch
```

3. Replace `np.uint8` with `np.int8` or `np.int64`: synthesis succeeds. `np.int16` fails like uint8.

(Also observed on the previous release: with `mqt.qecc==1.9.0` — whose plain pip install currently fails on newer toolchains because the optional qsample/projectq simulation dependency does not build, so a workaround install is needed — the same snippet does not raise; it keeps printing `Local minimum reached. Making matrix triangular.` and had not returned after 60 s, while the int8 version returns immediately.)

lsschmid:
Fixes #775.

Passing check matrices with a "wrong" integer dtype to `CSSCode` blew up later on. `uint8`/`int16` parity checks crashed the heuristic state-prep synthesis with a numba `No matching definition` error (the jitted scorer is only compiled for int8/int32/int64), and `bool` matrices were quietly mishandled because numpy's boolean `@` is a logical product, so the CSS orthogonality check computed the wrong result.

Rather than teach numba about every dtype, I normalize the check/logical matrices to int8 at construction — they're binary GF(2) arrays so it's lossless, and it keeps the whole code object consistent with the existing int8 annotations. I also widened the orthogonality check to an integer accumulator so `bool` inputs are validated correctly (`CheckMatrix` gets the same int8 normalization).

Added regression tests covering all the integer dtypes plus bool, and confirming genuinely non-orthogonal codes are still rejected.

### DIFF
--- a/src/mqt/qecc/codes/core/css_code.py
+++ b/src/mqt/qecc/codes/core/css_code.py
@@ -39,6 +39,10 @@ class CSSCode(StabilizerCode):
         Lz: npt.NDArray[np.int8] | None = None,  # ruff:ignore[invalid-argument-name]
     ) -> None:
         """Initialize the code."""
+        # Check/logical matrices are binary GF(2) arrays and are normalized to int8 below (via
+        # ``np.asarray``/``np.array``). This keeps the whole code object dtype-consistent with the
+        # int8 type annotations; without it, dtypes such as uint8/int16/bool propagate through
+        # ``symplectic``/``Hx``/``Hz`` and break numba-dispatched synthesis (see issue #775).
         if Hx is None and Hz is None:
             if n is None:
                 msg = "If no check matrices are provided, the code size must be specified."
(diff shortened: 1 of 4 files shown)


lsschmid:
@coderabbit review

PR #778, opened 2026-07-31
✨ Automated Equivalence Checking

MaxieHelenBichmann:
## Description

**Equivalence-checking methods for deciding permutation and local Clifford equivalence of stabilizer and CSS codes**, which I developed for my bachelor's thesis.

`are_permutation_equivalent`
`are_local_clifford_equivalent`
`is_local_clifford_equivalent_to_css`

Algorithms that practically check whether two codes define the *same code space up to a permutation or a local Clifford operation on the physical qubits*,  and provide a witness for those pairwise equivalences. Additionally supports checking whether a stabilizer code is local-Clifford-equivalent to an arbitrary CSS Code. They can be used in practice for codes with up to 30-50 physical qubits, depending on the specific equivalence problem.
> The algorithms are parameter-dependent hybrid procedures that rely on invariants, classical signature-based preprocessing, SAT encodings, brute-force search, graph-state methods, and matroid-based graph isomorphism.

Assisted-by:  GPT-5.6 Sol via Codex 

## Checklist

- [x] Any agent that created, edited, or submitted GitHub content was explicitly authorized for that scope, as required by our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/qecc/blob/main/docs/ai_usage.md).
- [ ] Every agent-authored or agent-edited public text body begins with the visible disclosure `🤖 *AI text below* 🤖` (titles are exempt).
- [x] I have disclosed AI assistance in the PR description.


lsschmid:
Thanks for the PR! Content looks fine, first I would like to settle some of the structural points.

**On your question (A):** I'd keep it as its own submodule. It keeps it clean and separate. otherwise there is too much clutter I would say.


In the following some general points from Claude, that sound reasonable:

## Duplication

For scale: 21.6% of the module's code lines (241 of 1117, blank/comments stripped) sit
inside a ≥6-line duplicated block.

- The 6-element local Clifford group is written out five times in five different
  representations (`apply_lc` L152, the two 60-line `z3.Implies` blocks at L541 and L661,
  `canonicalize` L470, and `_apply_lc_witness` in the tests) — please replace with one
  action table in a shared `_cliffords.py`.
- Both 60-line `z3.Implies` blocks are a 6-row table written as straight-line code; a
  single loop over that table replaces them.
- `_sat_stabilizer_code` and `_sat_css_code` in `permutation_eq.py` share the same encoding
  skeleton, including a 29-line verbatim clone (permutation variables, the two
  `_exactly_one` loops, model extraction) — extract `_encode_permutation`,
  `_encode_row_operations`, `_extract_permutation`.
- The "row operations" block appears five times across the two files.
- Inside `_sat_css_code`, the X and Z row-operation loops are the same logic written two
  different ways (list comprehension vs `for`/`append`).
- `_elementwise_map`, `_xor_list`, `_exactly_one` and `_reduce_stabilizer_generators` are
  copy-pasted between the two modules — move to a shared private module, keeping the
  CSS-aware version of the reduce.
- `_partition_columns_by_invariants` is defined twice in `permutation_eq.py`, differing
  only in annotation and sort key.
- The Gray-code weight-enumerator loop is duplicated between the CSS and stabilizer
  signature functions; only the weight function differs, so pass it as a parameter.

## Structure

- `_lse_stabilizer_code` is 329 lines with 17 nested functions and 4 levels of
  `def`-in-`def`; most are pure functions that should be module-level so they can be tested.
- The GF(4) linear algebra (`_gf4_rref`, `_gf4_row_basis`, `_gf2_gf4_matmul`) is defined
  inside a closure — it belongs in its own module next to `mod2`, with its own tests.
- `_extract_connected_components` is a hand-rolled BFS, but `networkx` is already a
  dependency and already imported in the sibling module (`nx.connected_components`).
- Rename the modules to full words to match the repo (`permutation_eq.py` →
  `permutation_equivalence.py`, same for the LC one).
- `LOCAL_CLIFFORDS` is defined at L183 under the "Decision procedures" banner but used at
  L168 under "Brute force algorithms" — move it to the top of the module.
- The dispatch thresholds 17, 20, 30 and `r < 10` are inline magic numbers while 5 is a
  named constant; name them all and note where they came from.

## Correctness and typing

- `_elementwise_map` uses `zip(..., strict=False)`, so a length mismatch silently
  under-constrains the SAT instance and yields a wrong "equivalent" verdict instead of an
  error — use `strict=True` (the repo's own `symbolic_vector_eq` raises here).
- Both signature functions return `tuple[bool, X | None, X | None]`, which forces four
  `assert ... is not None` calls that vanish under `python -O`; returning
  `tuple[dict, dict] | None` removes them.
- The module uses bare `np.ndarray` 22× vs `npt.NDArray[...]` 6×, inverted from the rest of
  the repo (351 vs 15) — and dtype matters a lot in this code.
- `preserved_low_degree_local_invariant` is the only helper without a leading underscore, so
  it will render as public API in the generated autoapi docs.

## Tests

- No test uses the built-in code constructions, unlike `tests/circuit_synthesis/`. this might make the tests a little less manual?! but not mandatory
- `_assert_maps_rowspace` and `_assert_maps_rowspace_stabilizer` have identical bodies with
  only the transform differing — one helper in a `conftest.py` maybe

## Docs and API

- **On your question (B):** the `Tools` toctree in `docs/index.md` contains only hub pages
  (`tools/synthesis`, `tools/decoders`, `tools/compilation`), each a grid of cards linking to
  the real feature pages; `EquivalenceChecking` is currently added there as a leaf page, so
  it needs a small hub of its own (e.g. `tools/analysis.md`) instead.
- None of the three functions are exported from the top-level `mqt.qecc` namespace, though
  the docs page starts with `from mqt.qecc import StabilizerCode`.
  
With that the PR has a good shape I would say. We can then also run coderabbit for further suggestions.

If you have any question, let me know!

MaxieHelenBichmann:
@lsschmid Thanks for the feedback, those are very valid points. I adapted the modular algorithms to have a more cohesive structure when considered as a whole, and reduced code duplication. The current state should be way better architecturally, and include the mentioned local fixes.

PR #690, opened 2026-05-14
Add function to apply CNOTs to a PureFaultSet

sunjerry019:
## Description

"Pushing" fault sets through a CNOT circuit is a commonly-used operation, implementing this function to a PureFaultSet seems like a good quality-of-life improvement. A PureFaultSet now also remembers what kind of error set it represents. 

Assisted-by: (Auto) via GitHub Copilot through inline suggestions and reviewed by me.

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [ ] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [ ] ~~I have added migration instructions to the upgrade guide (if needed).~~ not relevant
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [ ] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [x] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/qecc/blob/main/docs/ai_usage.md).
- [x] AI-assisted commits include an `Coauthored-by: [Model Name] via [Tool Name]` footer.
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

PR #783, opened 2026-08-03
Introduce XZFaultList for tracking X and Z faults simultaneously

sunjerry019:
## Description

Expands on the idea of `PureFaultSet` by having a list (i.e. stable order) of coupled $X$- and $Z$-Faults on which gates such as CNOTs, H, Reset may be applied. 

This PR is a fresh copy of the draft PR #712 without the edits to `PureFaultSet`, so that it can already be merged into main. Commits were picked from the branch XZFaultList.
In this case #690 need not be merged.

## Checklist

- [ ] ~I have updated the documentation to reflect these changes.~
- [ ] The changes follow the project's style guidelines and introduce no new warnings.
- [ ] The changes are fully tested and pass the CI checks.

- [x] AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.

### DIFF
--- a/src/mqt/qecc/circuit_synthesis/faults.py
+++ b/src/mqt/qecc/circuit_synthesis/faults.py
@@ -614,7 +614,7 @@ class XZFaultList:
         Raises:
             ValueError: If control or target indices are out of range or equal.
         """
-        self.ensure_apply_valid_input(control, target)
+        self._ensure_apply_valid_input(control, target)
 
         if inplace:
             # Apply CNOT directly to self.faults
@@ -650,7 +650,7 @@ class XZFaultList:
         Raises:
             ValueError: If qubit index is out of range.
         """
-        self.ensure_apply_valid_input(qubit)
+        self._ensure_apply_valid_input(qubit)
 
         if inplace:
             # Atomic swap using tuple assignment; use copies on RHS to avoid overlap
@@ -688,7 +688,7 @@ class XZFaultList:
         Raises:
             ValueError: If qubit index is out of range.
         """
-        self.ensure_apply_valid_input(qubit)
+        self._ensure_apply_valid_input(qubit)
 
         if inplace:
             self.faults["X"][:, qubit] = 0
@@ -733,7 +733,7 @@ class XZFaultList:
         # We do a simple logic, that every pair of X faults leads, in the worst case, to a Z fault on the other control. So we can just add all pairs of X faults as Z faults.
         # Z_i ^= (X_j & X_k) for all distinct i, j, k in {control1, control2, control3}
 
-        self.ensure_apply_valid_input(control1, control2, control3)
+        self._ensure_apply_valid_input(control1, control2, control3)
 
         if inplace:
             x_faults, z_faults = self.faults["X"], self.faults["Z"]
@@ -767,7 +767,7 @@ class XZFaultList:
             ValueError: If any qubit index is out of range.
             ValueError: If qubits are not distinct.
         """
-        self.ensure_apply_valid_input(control1, control2, target)
+        self._ensure_apply_valid_input(control1, control2, target)
 
         fault_list = self.apply_hadamard(target, inplace=inplace)
         fault_list.apply_ccz(control1, control2, target)
@@ -775,11 +775,11 @@ class XZFaultList:
 
         return fault_list
 
-    def ensure_apply_valid_input(self, *qubits: int) -> bool:
+    def _ensure_apply_valid_input(self, *qubits: int) -> None:
         """Ensures that the input into apply_* functions are valid.
 
         Returns:
-            bool: True if everything is okay
+            None: None if everything is okay
 
         Raises:
             ValueError: If any qubit index is out of range.
@@ -792,8 +792,6 @@ class XZFaultList:
         if n_q > 1 and len(set(qubits)) != n_q:
             msg = "All qubits must be different."
             raise ValueError(msg)
-
-        return True
 
     def reduce_to_coset_leaders(
         self, generators: tuple[npt.NDArray[np.int8] | None, npt.NDArray[np.int8] | None], inplace: bool = True
--- a/src/mqt/qecc/circuit_synthesis/faults.py
+++ b/src/mqt/qecc/circuit_synthesis/faults.py
@@ -835,5 +835,5 @@ class XZFaultList:
 
     def __repr__(self) -> str:
         """Return a string representation of the XZFaultList."""
-        repr_ = [object.__repr__(self), "X:", repr(self.faults["X"]), "Z:", repr(self.faults["Z"])]
+        repr_ = [object.__repr__(self) + f" num_qubits: {self.num_qubits}", "X:", repr(self.faults["X"]), "Z:", repr(self.faults["Z"])]
         return "\n".join(repr_)
--- a/src/mqt/qecc/circuit_synthesis/faults.py
+++ b/src/mqt/qecc/circuit_synthesis/faults.py
@@ -818,9 +818,9 @@ class XZFaultList:
         # use qecc_faults.coset_leader(single_fault, generators) for x and z
         ret = self if inplace else self.copy()
 
-        for error_type, g in zip(ret.faults, generators, strict=False):
+        for error_type, _g in zip(ret.faults, generators, strict=False):
             # Ensure generators are numpy arrays (may be empty)
-            g = None if g is None else np.asarray(g, dtype=np.int8)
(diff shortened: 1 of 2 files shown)


sunjerry019:
@coderabbitai full review

sunjerry019:
@lsschmid Could you look over this, and if it's okay, merge it? Thank you!

lsschmid:
I'll try to find the time the next days 👍

lsschmid (commented):
A few small things, mostly about staying consistent with `PureFaultSet`. Still thinking about the CCZ propagation model, will come back to that separately.

src/mqt/qecc/circuit_synthesis/faults.py
```diff
@@ -483,3 +486,380 @@ def t_distinct(fs1: PureFaultSet, fs2: PureFaultSet, t: int, stabs: npt.NDArray[
...
+            AssertionError: If generators tuple length is not 2.
+
+        Warning:
+            This calls :func:`coset_leader` once per fault row, so it might take a while.
+        """
+        # Setting the corresponding generator to None means no reduction is done
+
+        assert len(generators) == 2, "Generators should be a tuple of x_generators and z_generators"
+
+        logger.warning("This calls coset_leader once per fault row, so it might take a while.")
```
lsschmid, line 833: This fires on every call. The docstring already has a Warning section for it, so I'd either drop it or downgrade to `logger.debug`.

src/mqt/qecc/circuit_synthesis/faults.py
```diff
@@ -483,3 +486,380 @@ def t_distinct(fs1: PureFaultSet, fs2: PureFaultSet, t: int, stabs: npt.NDArray[
...
+
+        for error_type in ("X", "Z"):
+            g = generator_map[error_type]
+            if ret.faults[error_type].shape[0] > 0 and g is not None and g.size > 0:
+                for i in range(ret.faults[error_type].shape[0]):
+                    ret.faults[error_type][i] = np.asarray(coset_leader(ret.faults[error_type][i], g), dtype=np.int8)
+
+        return ret
+
+    def __repr__(self) -> str:
```
lsschmid, line 856: `object.__repr__` puts the memory address in here, so the output isn't stable across runs. `PureFaultSet.__repr__` just above returns a clean string, can we match that?

src/mqt/qecc/circuit_synthesis/faults.py
```diff
@@ -483,3 +486,380 @@ def t_distinct(fs1: PureFaultSet, fs2: PureFaultSet, t: int, stabs: npt.NDArray[
...
+            x_faults, z_faults = self.faults["X"], self.faults["Z"]
+            ret = self
+        else:
+            # Create a new XZFaultList with copied faults
+            new_list = XZFaultList(self.num_qubits)
+            new_list.faults["X"] = np.copy(self.faults["X"])
+            new_list.faults["Z"] = np.copy(self.faults["Z"])
+
+            x_faults, z_faults = new_list.faults["X"], new_list.faults["Z"]
+            ret = new_list
```
lsschmid, line 633: This copy-or-self block is repeated in `apply_hadamard`, `apply_reset` and `apply_ccz`. `reduce_to_coset_leaders` already does it as `ret = self if inplace else self.copy()`, would be nice to use that everywhere.

tests/circuit_synthesis/test_faults.py
```diff
@@ -637,3 +639,621 @@ def test_permute_qubits_inplace():
     fault_set.permute_qubits(permutation, inplace=True)
 
     assert fault_set != PureFaultSet.from_fault_array(faults), "Faults were not permuted correctly in place"
+
+
+"""XZFaultList Tests"""
```
lsschmid, line 644: This is a no-op expression, not a docstring. Make it a `#` comment.

src/mqt/qecc/circuit_synthesis/faults.py
```diff
@@ -483,3 +486,380 @@ def t_distinct(fs1: PureFaultSet, fs2: PureFaultSet, t: int, stabs: npt.NDArray[
...
+        """Initialise a XZFaultList object.
+
+        Args:
+            num_qubits (int): The number of qubits in the circuit
+        """
+        self.num_qubits = num_qubits
+        self.faults = {
+            "X": np.zeros((0, num_qubits), dtype=np.int8),
+            "Z": np.zeros((0, num_qubits), dtype=np.int8),
+        }
```
lsschmid, line 504: The `"X"`/`"Z"` keys are typed as a plain `dict[str, ...]`, so a typo like `faults["Y"]` only blows up at runtime. Two attributes (`x_faults`/`z_faults`, like the naming in `state_prep.py`) or a TypedDict would be safer. The tests also reach into `faults.faults["X"]` everywhere, so an accessor might be worth it.

src/mqt/qecc/circuit_synthesis/faults.py
```diff
@@ -483,3 +486,380 @@ def t_distinct(fs1: PureFaultSet, fs2: PureFaultSet, t: int, stabs: npt.NDArray[
...
+
+        Returns:
+            A new XZFaultList object with copied fault arrays.
+        """
+        new_list = XZFaultList(self.num_qubits)
+        new_list.faults["X"] = np.copy(self.faults["X"])
+        new_list.faults["Z"] = np.copy(self.faults["Z"])
+        return new_list
+
+    def __iter__(self) -> Iterator[tuple[npt.NDArray[np.int8], npt.NDArray[np.int8]]]:
```
lsschmid, line 594: There's `__iter__` but no `__len__`. Also no `__eq__`, `to_array` or `from_fault_array` like `PureFaultSet` has. `__eq__` especially will be needed as soon as something actually uses this class.

src/mqt/qecc/circuit_synthesis/faults.py
```diff
@@ -483,3 +486,380 @@ def t_distinct(fs1: PureFaultSet, fs2: PureFaultSet, t: int, stabs: npt.NDArray[
...
+        """Iterate over fault pairs in the list.
+
+        Yields:
+            Tuples of (x_fault, z_fault) for each row in the fault arrays.
+        """
+        for i in range(len(self.faults["X"])):
+            yield (self.faults["X"][i], self.faults["Z"][i])
+
+    def apply_cnot(self, control: int, target: int, inplace: bool = True) -> XZFaultList:
```
lsschmid, line 603: `inplace` defaults to `True` here, but `PureFaultSet.combine` defaults to `False`. Can we settle on one default for the module?

src/mqt/qecc/circuit_synthesis/faults.py
```diff
@@ -483,3 +486,380 @@ def t_distinct(fs1: PureFaultSet, fs2: PureFaultSet, t: int, stabs: npt.NDArray[
...
+        Args:
+            faults: A tuple of (x_fault, z_fault) where each is a 1D numpy array.
+                Each array must have length num_qubits.
+                One of the faults may be set to None, which is treated as an all-zero fault.
+
+        Raises:
+            ValueError: If fault arrays don't have the correct length.
+            ValueError: If both faults are None
+        """
+        assert len(faults) == 2, "Faults should be a tuple of x_fault and z_fault"
```
lsschmid, line 518: The type hint already fixes this to a 2-tuple, and asserts get stripped under `-O`. Same for the one in `reduce_to_coset_leaders`. Drop them, or raise a `ValueError` if you want the check to be real?

lsschmid:
Hey! In general this looks good to me, thanks a lot for putting it together.

One thing: I don't know exactly what you need this for, but for the CCZ model some things are unclear to me.


`z[c1] ^= x[c2] & x[c3]` says a Z only appears when *two* controls have an X
fault. But a single X already leaves a CZ behind (`CCZ·X₀·CCZ = X₀·CZ₁₂`), which
acts as a Z depending on the other qubit's state. The `&` only works if the
qubits start in |0…0⟩. But the prep circuits also use |+⟩, so a lone X can cause a Z that
this misses. If the |0…0⟩ assumption is intended, could you put it in the
docstring?

Second, the `^=` can cancel a real fault against an induced one: `x=(0,1,1),
z=(1,0,0)` comes out with no Z at all (line 941, you noticed it too). Fine for
exact Pauli tracking, but the docstring calls this adversarial, and there it
hides a fault, so maybe it should be `|=`?!

if that is clarified I guess we can merge

denialhaag:
Just FYI: The hanging Read the Docs build should fix itself the next time you push something. 🙂

PR #821, opened 2026-09-08
Remove LFS-tracked pkl files and drop LFS tracking rule + fixed broken link in docs

LSHerzog:
As the evaluation files in scripts `qecc/scripts/cococo/evaluations_movable_logical_qubits` where consuming the LFS bandwidth, I have removed them and stored them locally only. The results can be reproduced with the scripts in that repository. Furthermore, I found a broken link in my documentation.


burgholzer (approved):
Thanks for fixing this!

